PR #5, opened 2016-02-17
Setup Hello World and top level testing

parkerl:
I've been taking a stab at this. Getting testing going has been a bit...er...interesting. I will need some guidance setting up Travis. What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 .DS_Store
 bin/configlet
 bin/configlet.exe
+elm-stuff
 CHECKLIST
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,0 +1,13 @@
+for example_file in exercises/**/*.example
+do
+  exercise_dir=$(dirname $example_file)
+  exercise=$(basename $example_file .example)
+  echo 'setting up .....'
+  mv "$exercise_dir/$exercise.elm" "$exercise_dir/$exercise.impl"
+  mv "$exercise_dir/$exercise.example" "$exercise_dir/$exercise.elm"
+  echo 'building .....'
+  elm-test exercises/**/*Tests.elm
+  echo 'tearing down .....'
+  mv "$exercise_dir/$exercise.elm" "$exercise_dir/$exercise.example"
+  mv "$exercise_dir/$exercise.impl" "$exercise_dir/$exercise.elm"
+done
--- a/elm-package.json
+++ b/elm-package.json
@@ -1,0 +1,17 @@
+{
+    "version": "1.0.0",
+    "summary": "Exercism problems in Elm.",
+    "repository": "https://github.com/exercism/xelm.git",
+    "license": "BSD3",
+    "source-directories": [
+        ".",
+        "./exercises/hello_world"
+    ],
+    "exposed-modules": [],
+    "dependencies": {
+        "deadfoxygrandpa/elm-test": "3.0.1 <= v < 4.0.0",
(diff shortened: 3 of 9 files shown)


parkerl:
@kytrinyx had a chance to look at this? Do you think I'm on the right track?

kytrinyx:
Sorry, I just got back from Australia and the time confusion has stolen my brain. I'm way behind on all the exercism stuff. I'll be going through my backlog today.

.gitignore
```diff
@@ -2,4 +2,5 @@
 .DS_Store
 bin/configlet
 bin/configlet.exe
+elm-stuff
```
kytrinyx: :) Is this something that everyone will have to deal with, or could it go into `.git/info/exclude`?

bin/build.sh
```diff
@@ -0,0 +1,10 @@
+for example_file in exercises/**/*.example
+do
+  exercise_dir=$(dirname $example_file)
+  exercise=$(basename $example_file .example)
+  mv "$exercise_dir/$exercise.elm" "$exercise_dir/$exercise.impl"
+  mv "$exercise_dir/$exercise.example" "$exercise_dir/$exercise.elm"
+  elm-test exercises/**/*Tests.elm
+  mv "$exercise_dir/$exercise.elm" "$exercise_dir/$exercise.example"
+  mv "$exercise_dir/$exercise.impl" "$exercise_dir/$exercise.elm"
```
kytrinyx: I love love love the use of `.example` as the file extension, and rename to `.elm`. That's really clean!

elm-package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "1.0.0",
+    "summary": "helpful summary of your project, less than 80 characters",
+    "repository": "https://github.com/user/project.git",
```
kytrinyx: https://github.com/exercism/xelm.git

elm-package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "version": "1.0.0",
+    "summary": "helpful summary of your project, less than 80 characters",
```
kytrinyx: Exercism exercises in Elm.

kytrinyx:
Yes, this is great! I have a couple of tiny comments, but nothing important.

.gitignore
parkerl: Oddly this is where Elm puts dependencies and compiler crap. Seems that everyone will have their own. I would think this is the right place.

bin/build.sh
parkerl: Oh thanks, this whole copying around business seems hacky but it looks like the Haskell track is similar. Ugg

parkerl:
Awesome. I will make those updates and then see if we can get Travis off the ground?

.gitignore
kytrinyx: Huh. Yeah, sounds like we need it. `stuff`. So technical.

bin/build.sh
kytrinyx: And a bunch of other tracks, too.

kytrinyx:
Sounds great.

parkerl:
@kytrinyx Yikes...sorry for the churn...was trying to get Travis going :blush: 

Updates made. Travis build works and is green though it takes over 6 minutes which seems odd. Have you had build times like that with Travis on other projects?

https://travis-ci.org/parkerl/xelm/builds/111125214

kytrinyx:
Yeah, I've had some pretty slow build times. I don't know if it's particular steps of the installation, or just an artifact of running on shared machines.

PR #6, opened 2016-02-27
Add the Bob exerise

parkerl:
I think this is an ok start for Bob. I'm a bit unhappy with the random string generation. It turns out that Elm makes it a little difficult to generate random seeds. This implementation will always produce the same string of gibberish which is not ideal. However, there is so much scaffolding needed to make it truly random that I'm going to punt until someone with more Elm depth can take a look.

Let me know what else your eagle eye spots.


kytrinyx:
This seems like a great start. Thank you!

PR #22, opened 2016-03-12
Update README with dev workflow

parkerl:


README.md
```diff
@@ -36,6 +47,15 @@ Please keep the following in mind:
 
 - Each problem should have a test suite, an example solution, and a template file for the real implementation. The CI build expects files to be named using the following convention. The example solution should be named `ExerciseModuleName.example`. The template file should be named `ExerciseModuleName.elm`. Test file should be named `ExerciseModuleNameTest.elm`.
 
+- The recomended workflow when working on an exercise is to first create the implmentation and test files, `ExerciseModuleName.elm` and `ExerciseModuleNameTest.elm`.
```
tgecho: implmentation -> implementation

README.md
```diff
@@ -5,7 +5,18 @@ Exercism Exercises in Elm
...
+- Install `elm-test` for the command line with `npm install elm-test`.
+
+### Running Tests in an Exercise Directory
+Inside an exercise directory, Elm must be configured manually.
+
+- Install `laszlopandy/elm-console` and `deadfoxygrandpa/elm-test`.
+  - `elm-package install -y laszlopandy/elm-console`
+  - `elm-package install -y deadfoxygrandpa/elm-test`
+- You should then be able to run tests with `elm-test ExerciseModuleNameTest.elm`
+
```
tgecho: I've been running elm-test from the root of the repo, since it already had an `elm-package.json`. Since these requirements should be standard across all tests, might it make sense to have everyone do it that way rather then repeating these instructions for every test?

tgecho: As another bonus (and I just double checked this), the first time you run `elm-test` from the root it will ask for permission to install all of the elm packages specified in elm-package.json.

README.md
parkerl: How dare you correct my speeling! 

:wink:

README.md
parkerl: yeah, this is really more about how users will have to run the exercises once they have pulled one down. I think it maybe goes somewhere else eventually but thought it should go here as well.

tgecho: Yeah, I guess that would go in https://github.com/exercism/xelm/blob/master/docs/TESTS.md ?

I was trying to update that a bit more, but now I'm torn. Since we require elm-test and elm-console, users will need to reinstall both of them in every exercise folder. The least gross option I can think of is to include an elm-package.json file in each exercise directory so that the first run of elm-test will automatically install things for them. Is that kosher @kytrinyx @bewatts ?

parkerl:
@tgecho updated this to match your installation instructions in `INSTALLATION.md`. How's it looks?

README.md
```diff
@@ -36,6 +43,15 @@ Please keep the following in mind:
...
 - Each exercise should have a test suite, an example solution, a template file for the real implementation and an `elm-package.json` file with the `elm-test` and `elm-console` dependencies. The CI build expects files to be named using the following convention. The example solution should be named `ExerciseModuleName.example`. The template file should be named `ExerciseModuleName.elm`. Test file should be named `ExerciseModuleNameTest.elm`.
 
+- The recommended workflow when working on an exercise is to first create the implementation and test files, `ExerciseModuleName.elm` and `ExerciseModuleNameTest.elm`.
+  - Test the new exercise directly by running `elm-test exercises/exercise_module_name/ExerciseModuleNameTest.elm`. 
+
+  - Once the implementation of the exercise is complete, move `ExerciseModuleName.elm` to `ExerciseModuleName.example` and create the template file.
+
+  - Make sure everything is good to go by running all tests with `bin/build.sh`.
+
+- Please do not commit any Elm configuration files inside the exercise directories, `elm-package.json` or `elm-stuff`.
```
tgecho: This is actually no longer quite true... we actually *want* an elm-package.json in the exercise folder, but it should be the "standard" one.

Looks great otherwise :+1:

PR #35, opened 2016-03-14
Collect exercise failures rather than exiting early.

parkerl:
I generally feel that the build script should accumulate all the errors and report them before exiting. This is nice when testing changes to many exercises. However, we may want to fail fast on CI given the length of the build...thoughts?

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,3 +1,6 @@
+declare -i TEST_RESULT=0
+FAILED_EXERCISES=''
+
 for example_file in exercises/**/*.example
 do
   exercise_dir=$(dirname $example_file)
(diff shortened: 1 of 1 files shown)


tgecho:
I think you raise a good point. Looking at the Travis build time for this PR, I'd say the cache seems to be working. Pre-cache builds were taking 7-8 minutes or more, but now they seem to be holding steady just above 4 min. 

I support this PR.

bin/build.sh
```diff
@@ -7,12 +10,16 @@ do
   echo '-------------------------------------------------------'
   echo "Testing $exercise"
   elm-test $exercise_dir/*Tests.elm
-  TEST_RESULT=$?
+  if [ $? -ne 0 ]; then
+    TEST_RESULT=1
+    FAILED_EXERCISES+="$exercise\n"
```
tgecho: The `\n` is ending up in the console output, for example:
```
The following exercises failed:
Anagram\nBob\n
```

parkerl: Yup. WOMM but will update.

PR #38, opened 2016-03-16
Tweak initial exercise order based on subjective difficulty (closes #34)

### LINKED ISSUE #34: Initial exercise order
Now that we have a decent starting set of exercises (pending PR merges), I'm wondering about a good starting order. I've generally thrown them in haphazardly, somewhat based on the order in the other tracks I've pulled them from.

The current order, as of https://github.com/exercism/xelm/pull/31 is this:

1. hello-world
1. leap
1. pangram
1. rna-transcription
1. hamming
1. word-count
1. bob
1. run-length-encoding
1. difference-of-squares
1. anagram
1. raindrops
1. triangle

My questions for @kytrinyx, @parkerl and whoever else:

1. Is there a canonical/recommended order of exercises?
1. Is it a correct assumption that once launched it would be a bad idea to add new exercises anywhere other than the end?
1. Assuming 2 is correct, might it be worthwhile to sort the first batch of exercises by some subjective measure of difficulty to provide a nice onramp?

tgecho:
I took a very quick stab at an initial ordering. This is completely subjective based on glancing at each test/example for about 5 seconds and scribbling down the notable concepts from each. Obviously this is colored heavily by my own inexperience with the language.

hello-world
- string concatenation
- Maybe

bob
- case
- basic string

leap
- basic numbers

raindrops
- case, basic math/strings

pangram
- string manipulation
- List functions

triangle
- multiple if branches

anagram
- filtering, string manipulation

difference-of-squares
- basic numbers
- range syntax    

word-count
- Dict (possibly?)
- Regex (possibly?)

hamming
- string map2ing (in my case)

rna-transcription
- Result / err handling
- string mapping, folding, etc...

run-length-encoding
- relatively complicated map/fold

### DIFF
--- a/config.json
+++ b/config.json
@@ -6,17 +6,17 @@
   "test_pattern": "TODO",
   "problems": [
      "hello-world",
-     "leap",
-     "pangram",
-     "rna-transcription",
-     "hamming",
-     "word-count",
      "bob",
-     "run-length-encoding",
-     "difference-of-squares",
-     "anagram",
+     "leap",
      "raindrops",
-     "triangle"
+     "pangram",
+     "triangle",
+     "anagram",
+     "difference-of-squares",
+     "word-count",
+     "hamming",
+     "rna-transcription",
+     "run-length-encoding"
   ],
   "deprecated": [
 


kytrinyx:
This looks like a reasonable guess!

PR #42, opened 2016-03-19
Changes to exercises after an `elm-format`

parkerl:
This is what I get when I run `elm-format` on all the exercise examples. Look legit @tgecho ?


tgecho:
Yep, looks correct. I don't love what it did to RNATranscription or Raindrops... but I guess that's an issue to take up with elm-format.

parkerl:
Yeah, I figure if we push for folks to use it and then don't take what it gives us we will be fighting an uphill battle against PRs.

tgecho:
Agreed!

PR #43, opened 2016-03-19
Make it easy to run `elm-format` locally

parkerl:
I thought this might make it easy for folks to run `elm-format` before submitting a PR. Thought @tgecho ?

- run WITH_FORMAT=true bin/build.sh


tgecho:
Great idea! :+1:

PR #44, opened 2016-03-20
Standardize on module statements without `(..)`

tgecho:
This one is totally pedantic. I recently confirmed that the `(..)` is totally redundant in Elm. Both of these versions will export everything in the module:

```elm
module Foo (..) where
module Foo where
```

We are a bit inconsistent in our usage, though most of the exercises I added use the `(..)`. Since it's apparently superfluous, I think we should simply remove it.

### DIFF
--- a/exercises/accumulate/Accumulate.elm
+++ b/exercises/accumulate/Accumulate.elm
@@ -1,1 +1,1 @@
-module Accumulate (..) where
+module Accumulate where
--- a/exercises/accumulate/Accumulate.example
+++ b/exercises/accumulate/Accumulate.example
@@ -1,4 +1,4 @@
-module Accumulate (..) where
+module Accumulate where
 
 
 accumulate : (a -> b) -> List a -> List b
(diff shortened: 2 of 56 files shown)


parkerl:
@tgecho looks like elm-format puts these all back...?

tgecho:
That doesn't seem to be the case for me (or the CI build). Maybe an elm-format version mismatch?

```
$ elm-format --version
elm-format 0.2.0-alpha
```

tgecho:
Well, never mind again. I simply did a find and replace followed by a run of `./bin/build.sh`... but without the `ELM_FORMAT=1`. So yeah, you're right. Sorry about that. :)

parkerl:
@tgecho yeah sorry if it wasn't clear that running it is opt in. I thought about making it run by default locally. That would require contributors to install elm-format or they would get errors. Actually the build still runs, it just outputs some errors. Running on CI requires more configuration of Travis than I had time to tinker with yesterday. I don't think it's a bad idea though,

tgecho:
Nope, you were totally clear! I'm just jumping into this in fits and spurts, so I wasn't paying close enough attention when I created this PR. :)

I think running elm-format in the CI build is probably the right move if we're going to be strict about it going forward.

PR #46, opened 2016-03-20
Switch nucleotide-counts from `List (Char Int)` to `Dict Char Int`

tgecho:
I didn't look at this exercise hard enough when I picked it from the other repo. When I actually sat down to do my own from scratch solution, I realized that the `List (Char Int)` output it was looking for felt a bit arbitrary, with the counts needing to be in a specific (non alphabetical) order.

I revised it to look for a Dict instead, which seems like a more natural container for this sort of problem. Note that I did include a version, as recommended in https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#exercise-versioning

In the process I did learn something unfortunate about `Dict`: ["Dictionary equality with (==) is unreliable and should not be used."](http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Dict)

### DIFF
--- a/exercises/nucleotide-count/NucleotideCount.elm
+++ b/exercises/nucleotide-count/NucleotideCount.elm
@@ -1,1 +1,5 @@
 module NucleotideCount (..) where
+
+
+version =
+  2
--- a/exercises/nucleotide-count/NucleotideCount.example
+++ b/exercises/nucleotide-count/NucleotideCount.example
@@ -8,9 +8,13 @@ version =
   2
 
 
+empty =
+  Dict.fromList [ ( 'A', 0 ), ( 'C', 0 ), ( 'G', 0 ), ( 'T', 0 ) ]
+
+
 nucleotideCounts : String -> Dict.Dict Char Int
 nucleotideCounts dna =
-  String.foldl count Dict.empty dna
+  String.foldl count empty dna
 
 
 count : comparable -> Dict.Dict comparable Int -> Dict.Dict comparable Int
--- a/exercises/nucleotide-count/NucleotideCount.example
+++ b/exercises/nucleotide-count/NucleotideCount.example
@@ -1,18 +1,23 @@
-module NucleotideCount (..) where
+module NucleotideCount (nucleotideCounts, version) where
 
+import Dict
 import String
-import List
 
 
-nucleotideCounts : String -> List ( Char, Int )
-nucleotideCounts sequence =
-  [ (getCount 'A' sequence)
-  , (getCount 'T' sequence)
-  , (getCount 'C' sequence)
-  , (getCount 'G' sequence)
-  ]
+version =
+  2
 
 
-getCount : Char -> String -> ( Char, Int )
-getCount base sequence =
-  ( base, (List.length (String.split (String.fromChar base) sequence)) - 1 )
+nucleotideCounts : String -> Dict.Dict Char Int
+nucleotideCounts dna =
+  String.foldl count Dict.empty dna
+
+
+count : comparable -> Dict.Dict comparable Int -> Dict.Dict comparable Int
+count item counts =
+  Dict.update item incrMaybe counts
+
+
+incrMaybe : Maybe Int -> Maybe Int
+incrMaybe maybe =
+  (Maybe.withDefault 0 maybe) + 1 |> Just
--- a/exercises/nucleotide-count/NucleotideCountTests.elm
+++ b/exercises/nucleotide-count/NucleotideCountTests.elm
@@ -3,7 +3,8 @@ module Main (..) where
 import Task
 import Console
 import ElmTest exposing (..)
-import NucleotideCount exposing (nucleotideCounts)
+import NucleotideCount exposing (nucleotideCounts, version)
+import Dict
 
 
 tests : Test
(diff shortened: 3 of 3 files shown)


parkerl:
Let me apologize in advance for the longwinded response.

I love the change to `.example`. I'm going to wax a bit philosophic about the tests, so bear with me. Exercism is, of course, about learning. The exercises are intended to generate good conversations about code. With that in mind, I feel that tests should do their best to verify the correct output from the user's algorithm without being prescriptive of the implementation details used "under the hood". This often requires using more generic data structures in assertions. I think the exercises you have created so far have done a good job of this. (I have issues with some of the Elixir track exercises in this regard.)

Basically what I'm trying to say is changing the tests to require the use of a `Dict` takes the user out of the conversation.  They won't get to hear why that might be better than a `List`. Make sense?  `GradeSchool.allStudents` is similar. You could move `Dict.toList` into `nucleotideCounts`.

Sorry, not trying to be preachy. Feel free to disagree.

@kytrinyx love to hear your take on this.

tgecho:
No worries! I'm happy to more consciously reflect on what makes a good exercise.

I've viewed the test design as having two major parts: 1) verify a correct algorithm (as you say) and 2) teach/reinforce one or more idiomatic language concepts. Maybe I'm a bit off on the second one, though I think there's a lot of room to swing from one extreme to another.

In the case of the original exercise, my problem was that I think it *did* prescribe more about the implementation than I think is appropriate or helpful. The result list was required to be sorted in what appears to be an arbitrary order (A, T, C, G). Even the fact that the output is in an order sensitive data structure seems problematic to me.

To me, `Dict` seemed like the most appropriate and least prescriptive choice. I'd love to hear ideas for how a conversation about the merits of each could naturally fit into an exercise context, but nothing beyond a few sprinkled comments comes to mind.

All that said, my new test is a bit strange looking, what with its own `Dict.toList`. The goal was to allow the user to naively return a `Dict` without the burden of this little knowledge gem: ["Dictionary equality with (==) is unreliable and should not be used."](http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Dict). This could be exactly the wrong goal.

Beyond the fact that `Dict`s are not safely comparable, `toList` has the interesting (to me) property of promising to return a list sorted by key... thus my resulting test which accepts a `Dict` but still compares `List`s.

tgecho:
I also thought about using a `Record`, since the specific keys (er, fields in records?) is fixed, but that didn't feel quite right either. I'm afraid this is a case in which my Elm newbishness is making me uncertain.

kytrinyx:
I agree that exercises should be as non-prescriptive as possible (we have some exercises that aren't too awesome, because they have too many tests for arbitrary internal stuff. Over time I'm hoping to fix those.)

For sorting issues, how about sorting whatever you get back in the test so that it doesn't have to be ordered in any particular way in the solution?

parkerl:
OK so I totally agree that the ordering of the list in the test was arbitrary and needed to change. I also see your point that perhaps the exercise should demonstrate that a `Dict` is a more appropriate choice "out of the box". That said, I feel you should either:

1. continue to expect a list as the output and move the `Dict.toList` into `nucleotideCounts` as I suggest above. 

2. alter the tests to work better with `Dict` and remove implicit dependency on the ordering. Perhaps something like:

```
    , test
        "empty dna strand has no nucleotides"
        (assert
          (
           ((Dict.get 'A' (nucleotideCounts "")) == Just 0)
           && ((Dict.get 'C' (nucleotideCounts "")) == Just 0)
           && ((Dict.get 'G' (nucleotideCounts "")) == Just 0)
           && ((Dict.get 'T' (nucleotideCounts "")) == Just 0)
          )
        )
```

This gets a little verbose but could probably be refactored to look decent. 

As an additional comment, I don't agree with this change:

```
 -          [ ( 'A', 0 ), ( 'T', 0 ), ( 'C', 0 ), ( 'G', 0 ) ]
 -          (nucleotideCounts "")
 +          []
 +          (Dict.toList (nucleotideCounts ""))
```

I think that the implementation should still be required to return 0's.

tgecho:
I think we're nearly on the exact same page.

Right now the tests *are not* order dependent as far as the learner is concerned. In theory, `Dict`s are already order independent, but an outstanding problem makes Dict equality unsafe, [according to the docs](http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Dict) (I'm a little concerned about this, personally).

This is why I'm doing the `Dict.toList` on the test side: it turns out that the order *is* documented (the list is ordered by key). The learner only needs to return a `Dict`, and we create our own ordered and comparison safe `List`. It also means that they'll be blissfully unaware that there is a bug in Elm that will come back to bite them some day. :(

*If you think this is all too confusing and sketchy*, I'm really not stuck on it. I can totally live with the suggestion to simply sort the lists we receive by key, making order irrelevant. I still think a Dict is more appropriate, but it does come with a bit of baggage at the moment.

I'm also totally fine with requiring all four letters to be returned with default counts of 0 (commit added).

tgecho:
And... since I had some momentary bandwidth, here are two other options!

1. [Convert nucleotide-counts to use a record rather than a `List (Char, Int)`](https://github.com/exercism/xelm/pull/51)
2. [Sort returned nucleotide-count results to make the test order independent](https://github.com/exercism/xelm/pull/52)

I hope this doesn't feel like belaboring the point too much. I really do appreciate the back and forth thought on this.

parkerl:
@tgecho Sorry for the radio silence! Love to conversation on this. Let me take a look at the others!

parkerl:
Records FTW!

PR #50, opened 2016-03-22
Pin elm-console to 1.1.0 to fix a build failure introduced in 1.1.1

tgecho:


tgecho:
Well that's maddening. I seriously upgraded locally and everything broke. I then downgraded and it started working again. Poking a bit more...

tgecho:
So, this PR doesn't really seem to do anything. I reset the Travis cache (under settings) and now it seems to be passing :(

image

parkerl:
Huh. Irritating. Merge?

kytrinyx:
I'm confused--it started working again without pinning the build?

tgecho:
Nope. I cleared travis' build cache and it cleared up. I'm not sure if the issue is with Travis, elm-package or something we're doing. This PR seems to be a dead end. If this happens again I'll try to find some time to pin it down a little better.

kytrinyx:
Awesome, thanks for digging into it.

PR #51, opened 2016-03-24
Convert nucleotide-counts to use a record rather than a `List (Char, Int)`

tgecho:
As a spin off from https://github.com/exercism/xelm/pull/46, here's a different version using records. It's maybe a bit more prescriptive, but I actually kinda like it. Since there are actually only four possible keys, a fixed (and typed) record actually feels pretty reasonable. It also introduces records (which we haven't done) in a fairly simple way.

### DIFF
--- a/exercises/nucleotide-count/NucleotideCount.example
+++ b/exercises/nucleotide-count/NucleotideCount.example
@@ -1,18 +1,16 @@
 module NucleotideCount (..) where
 
 import String
-import List
 
 
-nucleotideCounts : String -> List ( Char, Int )
+nucleotideCounts : String -> { a : Int, t : Int, c : Int, g : Int }
 nucleotideCounts sequence =
-  [ (getCount 'A' sequence)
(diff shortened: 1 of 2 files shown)


exercises/nucleotide-count/NucleotideCount.example
```diff
@@ -1,18 +1,16 @@
...
-  ( base, (List.length (String.split (String.fromChar base) sequence)) - 1 )
+  let
+    getCount n =
+      String.filter (\c -> c == n) sequence |> String.length
+  in
+    { a = getCount 'A'
+    , t = getCount 'T'
+    , c = getCount 'C'
+    , g = getCount 'G'
+    }
```
tgecho: Record field names apparently need to start with a lower case letter :(

parkerl:
Ok I friggin' :heart: this one! I think a record makes total sense here. Nice job!!! :tada:

tgecho:
:+1: :tada:

PR #52, opened 2016-03-24
Sort returned nucleotide-count results to make the test order independent

tgecho:
As a spin off from https://github.com/exercism/xelm/pull/46, here's a simple version where we sort the output to make it more forgiving.


parkerl:
Records FTW!

PR #53, opened 2016-03-26
Add/increment nucleotide-count test version

tgecho:
I just realized that I forgot to add this to my records PR :(

Reference: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#exercise-versioning


parkerl:
Whoops I missed it too!

PR #55, opened 2016-03-27
Add series exercise

tgecho:
Two mildly notable things:

1. I changed the "overly long slice" to expect an empty list rather than an error, rationalizing that there simply aren't any runs of 4 in a 3 char string. I'm flexible /shrug
2. I added a test for input with non-numeric characters which expects an `Err`.

### DIFF
--- a/exercises/series/Series.example
+++ b/exercises/series/Series.example
@@ -1,0 +1,33 @@
+module Series (..) where
+
+import String
+import List
+import Result
+
+
+slices : Int -> String -> Result String (List (List Int))
+slices size input =
+  if size < 1 then
+    Err ("Invalid size: " ++ toString size)
+  else
+    String.split "" input
+      |> List.map String.toInt
+      |> combine
+      |> Result.map (takeRuns size)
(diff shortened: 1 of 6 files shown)


exercises/series/Series.example
```diff
@@ -0,0 +1,33 @@
...
+      |> Result.map (takeRuns size)
+
+
+takeRuns : Int -> List Int -> List (List Int)
+takeRuns size numbers =
+  let
+    candidate =
+      List.take size numbers
+  in
+    if List.length candidate < size || candidate == [] then
```
parkerl: `|| candidate == []` does not seem to be needed? Missing test?

tgecho: After a bit of head scratching, I now remember that it was to prevent a stack overflow before I put an explicit `size < 1` check at the very beginning of `slices`. I tweaked it to be a bit more explicit, but the defensive part of me would like to keep it :)

parkerl:
Awesome! :tada:

PR #56, opened 2016-03-27
Add additional test cases (including a property based test!) to run-length-encoding

tgecho:
A solution was submitted that seemed a bit off to me. After I whipped up a failing test case (which is added in this PR), I realized that this problem is a perfect candidate for property based testing.

I've never really written a property test beyond the hello world level before now, so I hope it doesn't look too crazy. It also seems to expose the same edge case failure that the original test suite missed. I did restrict the generated cases to sequences of upper case letters.

As for the actual execution, it does involve adding another dependency (elm-check). I wasn't sure if I wanted to make it an optional add-on, but for now it's imported as part of the regular tests. I incremented the version, though it should only be a breaking change for imperfect implementations! :)

### DIFF
--- a/exercises/run-length-encoding/RunLengthEncodingTests.elm
+++ b/exercises/run-length-encoding/RunLengthEncodingTests.elm
@@ -3,7 +3,8 @@ module Main (..) where
 import Task
 import Console
 import ElmTest exposing (..)
-import RunLengthEncoding exposing (decode, encode)
+import RunLengthEncoding exposing (version, decode, encode)
+import RunLengthEncodingPropertyChecks exposing (propertyTests)
 
 
 tests : Test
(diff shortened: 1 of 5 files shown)


exercises/run-length-encoding/RunLengthEncodingTests.elm
```diff
@@ -29,17 +33,24 @@ tests =
           (decode "12WB12W3B24WB")
         )
     , test
-        "decode(encode(...)) combination"
+        "(decode (encode (...)) combination"
```
parkerl: Good catch here. Personally I like to have this kind of housekeeping in a separate commit. It tends to add to the noise when doing code reviews (I'm a bit of a stickler about this in my day job but not so much in OS PRs). You can just commit and merge in yourself without review I would think.

parkerl:
@tgecho this is interesting. Normally I would say that this should be broken up into separate commits so he property based testing could be cleanly reverted if we decide we don't like it. Because this PR has been sitting here, I'm going to merge it in. 

As always, great work! :tada:

PR #57, opened 2016-03-28
Add robot-simulator exercise

tgecho:
I have to say I'm not too happy with my example. There are definitely some rough edges around updating nested records, and the other bits of repetition are bothersome.

### DIFF
--- a/config.json
+++ b/config.json
@@ -24,7 +24,8 @@
     "nucleotide-count",
     "phone-number",
     "grade-school",
-    "allergies"
+    "allergies",
+    "robot-simulator"
   ],
   "deprecated": [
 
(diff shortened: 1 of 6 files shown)


parkerl:
Ugg...this is my bad probably for being slow to review. Can you fix it up? I think it looks great! Go ahead and merge it when it is fixed. :tada:

tgecho:
No worries! We're bound to keep hitting these conflicts with parallel exercise PRs. Maybe we should take this suggestion? https://github.com/exercism/x-common/issues/216

kytrinyx:
I'm not actually sure that will help (almost certain that it won't). I don't think it's the comma, but the fact that two different PRs add an exercise next to the same entry.

PR #63, opened 2016-04-10
refactor nucleotide-count

lukewestby:
Hi!

I refactored nucleotide-count to accomplish the following goals:
- Reduce number of string iterations to 1
- Demonstrate fold(l|r|p)/update/init pattern which is common in Elm
- Demonstrate pattern-matching on literals
- Add a type alias for counts record

I know this is a significant change from the previous example but let me know what you think!

### DIFF
--- a/exercises/nucleotide-count/NucleotideCount.elm
+++ b/exercises/nucleotide-count/NucleotideCount.elm
@@ -1,5 +1,6 @@
 module NucleotideCount (..) where
 
 
+version : Int
 version =
   2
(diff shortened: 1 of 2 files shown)


tgecho:
Thanks!

I'll lead with the caveat that I'm new to this side of Exercism, so take it with a grain of salt when I say that my understanding of the purpose of the examples is to make sure the tests are sane and workable rather than to be a part of the end user experience.

That being said, these examples are my first mildly non trivial attempts at Elm, and I'd hate to have anyone assume they are idiomatic. So I'm not opposed to a bit of refactoring!

As for the PR itself, I suppose DNA can get pretty huge in real life, so I'll buy your optimization story :)

kytrinyx:
> I'm not opposed to a bit of refactoring!

Me neither; I just don't want people to agonize over the examples and end up holding up a launch (for example) because it's not *perfect*.

PR #64, opened 2016-04-10
use backward application operator

lukewestby:
This is a small change to replace `(\a r -> a r)` with the equivalent `(<|)`. This use of `(<|)` comes up occasionally in package sources and seeing it used in this way really helped me to understand it.

Thanks!

### DIFF
--- a/exercises/robot-simulator/RobotSimulator.example
+++ b/exercises/robot-simulator/RobotSimulator.example
@@ -95,4 +95,4 @@ simulate directions robot =
     directions
       |> String.toList
       |> List.map action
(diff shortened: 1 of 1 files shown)


tgecho:
First, this is cool and tweaked my perspective a bit, so thanks!

I'm not sure where it is on the spectrum of clear/too clever, so I'll come back and look again with a less pre-coffee brain.

lukewestby:
I think you're right on this being too much of a clever trick, when the original works perfectly well and is immediately clear. It was jarring to see the first time even though it did help. On second thought, this exercise is just right as it is :smile:

Thanks!

PR #65, opened 2016-04-10
refactor sublist

lukewestby:
Hi!

This refactor has one larger change and one minor one:
- Uses a union type to represent comparison possibilities between lists. Union types should always be favored over Strings or Ints for expressing finite option sets as it allows the compiler to enforce their usage/spelling/etc.
- `getLastInList` is a textbook use case for `Maybe.withDefault`

Let me know what you think!

### DIFF
--- a/exercises/sublist/Sublist.elm
+++ b/exercises/sublist/Sublist.elm
@@ -1,1 +1,8 @@
 module Sublist (..) where
+
+
+type ListComparison
+  = Equal
+  | Superlist
+  | Sublist
+  | Unequal
--- a/exercises/sublist/Sublist.elm
+++ b/exercises/sublist/Sublist.elm
@@ -1,8 +1,1 @@
 module Sublist (..) where
-
-
-type ListComparison
-  = Equal
-  | Superlist
-  | Sublist
-  | Unequal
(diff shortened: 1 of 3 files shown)


tgecho:
+10 on the union type! I would actually go so far as to remove them from `Sublist.elm`, as this is a relatively late exercise and one could argue that creating the type is part of the exercise.

You got me on `Maybe.withDefault`. Actually, I ended up gutting this entire function for my final solution and ended up with a `Maybe` free version:

```
inList : List a -> List a -> Bool
inList needle haystack =
  if needle == take (length needle) haystack then
    True
  else if length needle < length haystack then
    inList needle (drop 1 haystack)
  else
    False
```

tgecho:
Almost forgot! Because this is a breaking change, be sure to add a version to the example/stub and an accompanying check in the test.

parkerl:
:+1: Please add the version as @tgecho recommends. 

There is an example of adding versioning here https://github.com/exercism/xelm/commit/03be35553d002c0c28d24059eb7be62b951eff77

Thanks!

lukewestby:
Version added!

tgecho:
Thanks!

PR #70, opened 2016-04-17
Add shebang to `bin/build.sh`

NobbZ:
Without the shebang, the buildscript might fail on environments which do
not use `bash` or a compatible shell as default.

This PR fixes the issue and makes the script more robust to the
environment.


tgecho:
Good catch. Thanks!

PR #71, opened 2016-04-18
Add list-ops exercise

jcshih:
I'm unsure about adding large list tests, as elm doesn't have TCO, and it isn't really elm's intention to do large computations. Thoughts?

### DIFF
--- a/config.json
+++ b/config.json
@@ -26,7 +26,8 @@
     "phone-number",
     "grade-school",
     "allergies",
-    "robot-simulator"
+    "robot-simulator",
+    "list-ops"
   ],
   "deprecated": [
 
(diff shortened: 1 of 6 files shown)


tgecho:
This is great! My only hesitation about just merging is this issue: https://github.com/exercism/x-common/issues/129

Accumulate and strain both kind of duplicate a subset of this test, and it looks like a rename from `list-ops` to `sequence-ops` is on the table. I think we can probably merge anyway, but I'd like let it sit for a day or so.

jcshih:
Thanks! No problem. I don't really understand the exact relationship of the same exercise across languages, so I defer to you and others on this decision.

kytrinyx:
Problem descriptions / metadata are language-independent. Then any of the languages on Exercism can add their own implementation (test suite + reference solution). The README for the exercise is a combination of language-independent data and language/implementation specific hints.

tgecho:
Thanks @jcshih 🎉

PR #72, opened 2016-04-25
Fixed order of parameters for Hamming exercise's tests.

RiderSargent:
This fixes the order of parameters in the tests for the Hamming exercise.

The `assertEqual` function in `elm-test` considers the first parameter to be the "expected" value and the second parameter to be the "got" value, which is the reverse of how these tests were written. Hopefully this will make `elm-test`'s output more helpful and clear.

Any and all suggestions for making this better are welcome! Thanks.


parkerl:
Thanks @RiderSargent 🎉

PR #73, opened 2016-04-29
Add atbash-cipher exercise

tgecho:


parkerl:
@tgecho build failure?

tgecho:
Eh, I messed up the rebase on master. I also need to figure out why I'm not getting failure notifications from Travis.

kytrinyx:
It seems like per the documentation you should be receiving notifications. Weird. https://docs.travis-ci.com/user/notifications/#How-is-the-build-email-receiver-determined%3F

tgecho:
Hrm, it seems that emails can't be triggered for PR build failures.
> Pull Request builds do not trigger email notifications.

It's been a while since I worked with a Travis based project, so I guess I misremembered how this worked.

Oh well, guess I'll just need to be better about checking in. :)

kytrinyx:
There's always the explicit approach:

```
notifications:
  email:
    recipients:
      - one@example.com
      - other@example.com
    on_success: [always|never|change] # default: change
    on_failure: [always|never|change] # default: always
```

exercises/atbash-cipher/AtbashCipher.example
```diff
@@ -0,0 +1,63 @@
...
+encode : String -> String
+encode plain =
+  let
+    translate =
+      toTranslator alphabet tebahpla
+  in
+    plain
+      |> String.toLower
+      |> String.filter (\c -> Char.isLower c || Char.isDigit c)
+      |> String.map translate
```
lukewestby: lovely use of `String.map` 😄

exercises/atbash-cipher/AtbashCipher.example
```diff
@@ -0,0 +1,63 @@
...
+
+
+alphabet : String
+alphabet =
+  "abcdefghijklmnopqrstuvwxyz"
+
+
+tebahpla : String
+tebahpla =
```
lukewestby: i personally enjoy this variable name very much but it's more "the Elm way" to call it `reversedAlphabet` or something of that nature

exercises/atbash-cipher/AtbashCipher.example
```diff
@@ -0,0 +1,63 @@
...
+  String.reverse alphabet
+
+
+toTranslator : String -> String -> Char -> Char
+toTranslator from to =
```
lukewestby: this is a super neat way to solve this problem, i love it!

exercises/atbash-cipher/AtbashCipher.example
tgecho: I've already come halfway. In an early (12something AM) version it was actually a "tebahpla : Gnirts", but I figured the type alias was a bit much. :)

You are of course completely correct.

tgecho:
Thanks @lukewestby!

PR #75, opened 2016-05-13
Update to support Elm 0.17

### LINKED ISSUE #74: Upgrade exercises to Elm 0.17
The only real code change that should be necessary is updating the module lines from `module Main (..) where` to `module Main exposing (..)` form. We're not doing any signal stuff yet, so the rest of the upgrade doc shouldn't affect us: https://github.com/elm-lang/elm-platform/blob/master/upgrade-docs/0.17.md

It also looks like we may need to switch to `elm-community/elm-test` to gain 0.17 support. I'm traveling with sketchy internet access at the moment, so I won't be able to sort this out in the next day or two.

tgecho:
Aimed to close #74

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bin/configlet
 bin/configlet.exe
 elm-stuff
 CHECKLIST
+build.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ sudo: false
 install:
   - nvm install 0.12
   - nvm use 0.12
-  - npm install -g elm@0.16.0 elm-test@0.16.0
+  - npm install -g elm@0.17.0 elm-test@0.16.1-alpha4
   - elm package install -y
 
 script:
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Please keep the following in mind:
 - Each exercise should have a test suite, an example solution, a template file for the real implementation and an `elm-package.json` file with the `elm-test` and `elm-console` dependencies. The CI build expects files to be named using the following convention. The example solution should be named `ExerciseModuleName.example`. The template file should be named `ExerciseModuleName.elm`. Test file should be named `ExerciseModuleNameTest.elm`.
 
 - The recommended workflow when working on an exercise is to first create the implementation and test files, `ExerciseModuleName.elm` and `ExerciseModuleNameTest.elm`.
-  - Test the new exercise directly by running `elm-test exercises/exercise_module_name/ExerciseModuleNameTest.elm`. 
+  - Test the new exercise directly by running `elm-test exercises/exercise_module_name/ExerciseModuleNameTest.elm`.
 
   - Once the implementation of the exercise is complete, move `ExerciseModuleName.elm` to `ExerciseModuleName.example` and create the template file.
 
@@ -57,10 +57,8 @@ Please keep the following in mind:
 - Test files should use the following format:
 
 ```elm
-module Main (..) where
+module Main exposing (..)
 
-import Task
-import Console
 import ElmTest exposing (..)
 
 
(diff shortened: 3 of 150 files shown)


tgecho:
@lukewestby I've got it almost there, but I get the following output when a test "succeeds":

```
Success! Compiled 1 module.
Successfully generated /var/folders/c8/lwxn19p55rj7mq0n7xhj0hv00000gp/T/elm_test_116412-8146-w6s0yx.js
Successfully compiled HelloWorldTests.elm
Running tests...
  1 suites run, containing 3 tests
  All tests passed


: ""

[stdin]:7348
    if (typeof Elm === "undefined") { throw "elm-io config error: Elm is not defined. Make sure you call elm-io with a real Elm output file"}
                                      ^
elm-io config error: Elm is not defined. Make sure you call elm-io with a real Elm output file
```

elm version 0.17.0
elm-test version 0.16.1-alpha4

tgecho:
Also, the sublist example overflows the stack... not sure why not and not in 0.16, but I don't have time to fix it tonight.

Everything else seems fine!

tgecho:
So `elm-make *Tests.elm --output tests.js && node tests.js` works. Apart from the fact that this is quite a bit longer than the old version (which is fixable with a trivial wrapper), ideally I'd like to not change the test running instructions on people. Any ideas @parkerl @lukewestby ?

lukewestby:
I can check with Richard to see if node-elm-test is going to get an update for 0.17 and get back to y'all today

lukewestby:
So from what I can gather node-elm-test is currently undergoing improvements and will be back at some point. My recommendation would be to move the actual test command into a separate shell script anyway to mitigate against things going into flux during future updates to Elm or to elm-test. Then the actual test running process could be refactored at will without changing the user experience.

tgecho:
I just added a simple wrapper script to each exercise (.sh for *nix and .bat for M$ 😄). It uses a wildcard so it can simply be copied into each exercise directory. Assuming everyone is happy with this approach (@parkerl?) , the only thing left is to fix the Sublist stack overflow.

I'll do that later tonight assuming no one else sends me something.

lukewestby:
Nice!

Re sublist:
The sublist problem is blowing out the stack on the call to `List.take` in `inList`. `List.take` in core isn't tail-recursive so we're not getting any help from the compiler and doing some tricks to get a trampoline in place even if possible is going to be immensely non-performant. Trampolining might not be something y'all want to expose force the user to learn about here anyway. Will keep trying and keep y'all posted.

tgecho:
@lukewestby I just figured that out myself... and maybe overeagerly submitted a bug to core (https://github.com/elm-lang/core/issues/601), since a simple `List.take 5707 [0..10000]` triggers the overflow.

We can decide how important huge lists are to this exercise, but in the meantime I bumped down the size of the lists in the offending test.

lukewestby:
sounds good! i have an idea using `foldl` that could work that i'll run past you just to see what you think but just reducing the size of the list seems like a great way to go

lukewestby:
Nevermind, it's unbelievably slow 😛

tgecho:
I'm game for whatever. Feel free to PR my PR branch (or do it as a followup). Either way I'd like to give @parkerl a bit of chance to glance at this branch before we merge. Or if you've tested out a few exercises locally yourself I can just merge it tomorrow. After that we'll still need to wait for x-api to get updated, so I don't want to delay too much!

lukewestby:
I think changing the length of the list for sublist is a good way to go, and I hope that issue you've put on core results in some good discussion 😄 

This all looks excellent! Thank you so much for spending time working on this it's such a great resource.

The one and only change I'd like to suggest to this PR is adding `--yes` to the line `elm-make $exercise_dir/*Tests.elm --output build.js && node build.js` in the build script so it skips the interactive bit.

tgecho:
Ah, good catch. Packages should already be installed by the `.travis.yml`, but this doesn't hurt things.

tgecho:
This will go live next time x-api is updated. Thanks everyone!

kytrinyx:
Deployed.

tgecho:
Thanks!

PR #84, opened 2016-06-18
Add say exercise

tgecho:
I'm not totally happy with the clarity of the example for this one, but it all seems to work. We'll see what everyone else does with it!

@parkerl

### DIFF
--- a/exercises/say/Say.example
+++ b/exercises/say/Say.example
@@ -1,0 +1,126 @@
+module Say exposing (say, SayError(..))
+
+import Array
+import Dict
+
+
+type SayError
+    = Negative
+    | TooLarge
+
+
+say : Int -> Result SayError String
+say number =
+    if number < 0 then
+        Err Negative
+    else if number < 20 then
+        Ok (saySmall number)
+    else if number < 100 then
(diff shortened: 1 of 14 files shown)


exercises/say/Say.example
```diff
@@ -0,0 +1,131 @@
...
+    if number < 0 then
+        Err Negative
+    else if number < 20 then
+        Ok (saySmall number)
+    else if number < 100 then
+        if number % 10 == 0 then
+            Ok (saySmall number)
+        else
+            Ok (sayTens (number // 10 * 10) ++ "-" ++ (saySmall (number % 10)))
+    else if number < k then
```
lukewestby: i think it would be okay to hardcode each of these constants

lukewestby:
Looks great!

exercises/say/Say.example
tgecho: The `k`, `t`, etc...? I also thought about expanding them to at least full words.

lukewestby: full words would be good too. it took me a second to figure out that they were abbreviations for the SI prefixes

PR #86, opened 2016-06-19
Add largest-series-product exercise

### LINKED ISSUE #85: WIP: largest-series-product

tgecho:
Closes #85

### DIFF
--- a/exercises/largest-series-product/LargestSeriesProduct.example
+++ b/exercises/largest-series-product/LargestSeriesProduct.example
@@ -1,0 +1,53 @@
+module LargestSeriesProduct exposing (largestProduct)
+
+import String
+
+
+largestProduct : Int -> String -> Maybe Int
+largestProduct length series =
+    if length < 0 then
+        Nothing
+    else if length == 0 then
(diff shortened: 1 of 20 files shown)


exercises/largest-series-product/LargestSeriesProduct.example
```diff
@@ -0,0 +1,53 @@
...
+
+largestProduct : Int -> String -> Maybe Int
+largestProduct length series =
+    if length < 0 then
+        Nothing
+    else if length == 0 then
+        Just 1
+    else
+        String.toList series
+            |> List.map (String.fromChar >> String.toInt)
```
tgecho: That's really really tempting, but unfortunately we're supposed to return `Nothing` if there is a non numeric char in the series, so we can't just discard them. So this would require a pre-validation step (which I'm not totally opposed to).

I will at least take the `String.split ""` simplification. :)

lukewestby: oh no! curse my poor reading skills! sounds good.

PR #87, opened 2016-06-19
Run pre Elm 0.17 exercises through elm-format version 0.3.0-alpha

tgecho:
Just a bit of housecleaning...


tgecho:
I'll merge in a day or so if no one objects.

Issue #47, opened 2016-03-20
Run elm-format on CI


parkerl:
Because we are being strict about formatting we should fail the build if elm-format causes diffs.

justgage:
I would be a little afraid that this would cause people to become a little discouraged. I think suggesting to use it with a bot in a comment after they submit it could be a good thing.

kytrinyx:
@justgage To clarify this about pull requests on GitHub to the Exercism Elm track, not about failing the tests when people are submitting exercises to the website.

justgage:
Ah yes, that's a good idea. Sorry about that.

kytrinyx:
No apologies necessary—you've got everyone's best interest in mind (and I agree with you: this would be frustrating and potentially quite discouraging if we did it on the website).

PR #96, opened 2016-08-14
Update to elm-test 2.0

### LINKED ISSUE #76: runtests.sh is not downloaded with executable permissions
```
$ ./runtests.sh
-bash: ./runtests.sh: Permission denied
```

I didn't even think to check/ask, but it looks like this behavior was changed a few months ago: https://github.com/exercism/cli/issues/276 :(

So... seeing as in my haste I've failed to create a clean/generic wrapper on the first try, I see a few obvious options:

1. Remove the wrappers, switch the instructions back and wait for an updated `elm-test` to ship. `elm-test MyTests.elm` actually works right now, it just has a bunch of extraneous garbage at the end of the output.
2. Update the instructions to `bash runtests.sh`.
3. Write a node.js based wrapper (which could actually be cross platform) and instruct them to run `node runtests.js`.

Thoughts @parkerl @lukewestby?

### LINKED ISSUE #95: Update to use the latest version of Elm Test
Hello!

Currently the Elm exercises are using an older version of Elm Test that is unmaintained and uses ugly hacks to report tests results to the user.
I believe we should update to use the latest version.

There are a few problems with this:
- The current version has a new syntax for tests that we would need to update to.
- Due to the nature of Elm there isn't an obvious way to do IO, so there are multiple solutions for returning test results to the user. We would need to decide how to run out tests in a fashion that is user friendly and does not pull too many dependencies into the exercises.

tgecho:
This is an example of what an exercise upgraded to elm-test 2.0 looks like (fixes #95). Among other highlights, I've reenabled the node elm-test runner workflow (`npm install -g elm-test`) as before, so the output is much nicer and by ditching the runtests scripts we can close #76.

- [x] Update the rest of the exercises
- [x] Update the installation/running docs
- [x] Get a 👍  from at least one other human being

### DIFF
--- a/exercises/accumulate/AccumulateTests.elm
+++ b/exercises/accumulate/AccumulateTests.elm
@@ -1,6 +1,9 @@
-module Main exposing (..)
+port module Main exposing (..)
 
-import ElmTest exposing (..)
+import Test.Runner.Node exposing (run)
+import Json.Encode exposing (Value)
+import Test exposing (..)
+import Expect
 import Accumulate exposing (accumulate)
 import String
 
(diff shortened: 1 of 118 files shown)


exercises/accumulate/AccumulateTests.elm
```diff
@@ -12,18 +15,21 @@ square x =
 
 tests : Test
 tests =
-    suite "Accumulate"
+    describe "Accumulate"
         [ test "[]] Accumulate"
-            (assertEqual [] (accumulate square []))
+            (\() -> Expect.equal [] (accumulate square []))
```
lpil: I think this syntax is the convention the Elm Test community are trying to promote:

```elm
test "square Accumulate" <|
    \() -> 
        (accumulate square [ 1, 2, 3 ])
            |> Expect.equal [ 1, 4, 9 ] 
```

lpil:
I like where this is going so far :)

Do we run Elm Format on the exercises?

exercises/accumulate/AccumulateTests.elm
tgecho: Yeah... To be honest I was able to do 99% of this via regex flavored find/replace because I was keeping the existing closing paren at the end of each test. Switching to that style will require a bit more work. That's really my strongest objection :)

tgecho:
Not in CI. There's an issue for that (#47), but no one has bothered to try installing it in Travis yet.

exercises/accumulate/AccumulateTests.elm
lpil: I think we should do it, but perhaps it's something we can leave for a later pull request, or just do incrementally.

tgecho: Ok, you got me. I figured how to get 95% of it in bulk, and elm-format helped me sort out the  rest pretty quickly.

tgecho:
Anyone have any objections to merging this (squashed, etc...)? @lpil @parkerl

lpil:
Looks great! Thank you for doing all this work.

PR #97, opened 2016-08-14
Use a type to express triagles instead of strings

lpil:
Hello!

For enum values such as this one found in the Triangle exercise I think we should avoid recommending use of strings in place of real types.

### DIFF
--- a/exercises/triangle/Triangle.elm
+++ b/exercises/triangle/Triangle.elm
@@ -1,1 +1,6 @@
 module Triangle exposing (..)
+
+
+version : Int
+version =
+    2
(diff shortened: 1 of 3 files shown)


tgecho:
This makes sense. We try to version our tests when we make a breaking change. Can you set the version of this test along the lines of [this NucleotideCount example](https://github.com/exercism/xelm/blob/cf3bd4da4768bde5e6aa995b3dacb57c201e0e12/exercises/nucleotide-count/NucleotideCountTests.elm#L10)?

lpil:
There you go :)

tgecho:
Thanks!

PR #103, opened 2016-08-21
Correct docs

lpil:
This brings the docs into line with the current test format


tgecho:
Ah, good catch. Though to be a correct example, the exposing/imports at the top will need to be updated as well.

tgecho:
Thanks!

PR #105, opened 2016-08-21
Upgrade Node Test Runner and Test Files

jehoshua02:
Fixes https://github.com/exercism/xelm/issues/100

Test

```
rm -fr ./elm-stuff ./**/*/elm-stuff
sh ./bin/build.sh
```

All tests should pass without errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ tests =
     ]
 
 
-main : Program Never
+main : Program Value
 main =
   runSuite tests
 ```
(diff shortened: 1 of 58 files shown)


jehoshua02:
CI failed *sigh*

I was able to reproduce locally by matching my environment with [ci config](https://travis-ci.org/exercism/xelm/jobs/154014106/config).

I'll look into it when I have time. Probably Monday/Tuesday. Somebody else might get to it faster than I can.

jehoshua02:
I tried node 6, 5, 4.5 and they fail. But for some reason node 4.2.1 passes. Strange.

tgecho:
This seems to be exactly the correct solution. I've been having a bit of trouble with Elm's cached compile output, so I'm rerunning it now.

tgecho:
It looks like a few more pieces needed to get synced up to make this all pass. I've added a new commit, which required a new PR: https://github.com/exercism/xelm/pull/106

PR #113, opened 2016-08-26
Run Elm Test in CI

jehoshua02:
Resolves https://github.com/exercism/xelm/issues/47

Feedback welcome.

image


tgecho:
Oh, nice, and you even got it down to one run for the whole build!

PR #124, opened 2016-12-15
Implements the bowling kata

chadfennell:
Example :bowling: implementation is based upon the [canonical list of tests](https://github.com/exercism/x-common/blob/master/exercises/bowling/canonical-data.json).

I'm sorta new to Elm. A few thoughts I had after working on this kata:

- scoreFrames has a whole bunch of let expression variables.
- I'm making multiple passes through the data in different ways. Maybe there ways to avoid that? Better use of pipe function application?
- The union Frame type is very simple. Perhaps it could somehow implicitly handle some of the validations?

Thanks!


tgecho:
Thanks for the PR!

You wouldn't have gotten a notification, but I'm afraid the CI build didn't work. You'll need to add an entry to the root `elm-package.json` and `config.json` as well as rename `Bowling.elm.example` to `Bowling.example`. I'll try to check it out locally tomorrow.

Thanks again!

chadfennell:
Oops, will do. Thanks!

chadfennell:
Looks like I also need to adjust the testing format. I'll change that as well.

chadfennell:
I gave elmenv a go in order to test a downgraded version of Elm but elmenv failed on 0.17.1 for me (on Ubuntu). 

But it got me wondering - would you mind if I simply updated the dependencies to 0.18 for the core repo and all of the current exercises prior to PRing for the bowling kata? I suspect others will run into similar issues, and it seems simpler to point them to an upgrade path rather than a downgrade path.

But maybe I'm missing something too.

Thanks.

tgecho:
@chadfennell We've had some discussion over at https://github.com/exercism/xelm/issues/122 ... I personally haven't had much time to think/work on it. Agree that upgrading to 0.18 would be ideal.

chadfennell:
@tgecho cool. I was being a bit thick too - just looked at the build script and, duh, I can install v 17 via npm. I'm making my test compliant with the current setup and will chime in on the other thread / help out there.

tgecho:
@chadfennell Sorry I let this get stale! We updated to Elm 0.18 a while back, and recently sorted out compatibility with the latest version of elm-test (0.18.3). Most of the updates should be pretty mechanical.

For the exercise itself, it looks like the [problem description](https://github.com/exercism/x-common/blob/master/exercises/bowling/description.md) is a bit specific. They want a separate roll/score operations, while you've combined everything into a single roll function.

I think we should match up with this using signatures along the lines of

```
roll : Int -> game -> game
score : game -> Int
```

This will allow the user to define their own game type. Your test file would just need a simple `List.foldl` based wrapper like

```
scoreGame : List Int -> Int
scoreGame scores =
    List.foldl Bowling.roll 0 scores
        |> Bowling.score
```

Just let me know if you don't have time and I'll take a stab at finishing this up.

PR #126, opened 2016-12-17
Switch to an npm managed setup + upgrade to Elm 0.18

tgecho:
This is a first stab at switching the tests to using a local `package.json` for each exercise. This enables the user to use a copy of elm/elm-test specific to that exercise. This should give us more flexibility around versioning and is just more hygienic overall.

**Background:**
- [Upgrading to Elm 0.18](https://github.com/exercism/xelm/issues/122)

**Highlights:**
- AFAICT the contents of SETUP.md will be appended to the README.md that is fetched with each exercise. This should provide up to date install/test instructions even if they've previously had old versions of previous exercises.
- I'm not sure I want to muddy the waters with mentioning yarn, but I'm tempted to leave a yarn.lock file for anyone who cares to use it. `yarn install` and `yarn test` seem to work just fine.
- I renamed the test and example files to simple `Tests.elm` and `Example.elm`. This makes it easier to ditch the wildcard in the test command (better for windows) and enables syntax highlighting in the example file.

**Known todos:**
- [x] Get some feedback
- [x] Test on windows (I think it'll work)
- [x] Test on linux (I'm pretty sure it'll work)
- [x] Migrate the rest of the exercises
- [x] Update the central elm docs/build.sh

CC: @fpapado @chadfennell @ryantriangles @rebelwarrior


chadfennell:
I like the idea of supporting multiple versions in order to make it easier to upgrade exercises incrementally when a new release occurs. 

>  Tests.elm and Example.elm

Yes, for the reasons you mentioned. I mistakenly submitted my first kata in this format, seems like a natural default.

- [x] Test on linux 

Works on Ubuntu 16.04.

tgecho:
Ok, looks like the central build part is working. I had to back off the example rename a bit in lieu of coming up with a more clever way to do the detection in `build.sh`, but it's still an improvement.

tgecho:
Should be good to go. I'll let it sit overnight in case anyone has an last minute thoughts/objections (and until I have a chance to do a smoke test in windows).

tgecho:
Windows seems 👍

fpapado:
Hey @tgecho, great work! Didn't get to reply on the earlier gb issue, I agree that this seems the best approach.

PR #128, opened 2016-12-27
Improve pangram test

duckmole:
* Add sentence missing 'z'

My first submit was incorrect but pass all the test : [my first solution](http://exercism.io/submissions/4cd7170d66a540bcbd61894081bccbd0)
I propose to add a test where a 'z' is missing. I try to find (google) an English sentence where a 'z' is missing but I didn't succeed.

Could you improve my proposal with an English sentence


tgecho:
Thanks!

PR #149, opened 2017-06-18
Remove leading 'x' from repository name

kytrinyx:
The leading 'x' is kind of arbitrary. Especially now that we can set
topics on the repositories, we don't need a pattern to distinguish what
is a track or not.

The repository itself has already been renamed. GitHub redirects from
the old name to the new name, so we do not have to rush to fix links to
the old repository name, though we should update them for the sake of
clarity.

See exercism/meta#1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-# xElm
-[![build status](https://travis-ci.org/exercism/xelm.svg?branch=master)](https://travis-ci.org/exercism/xelm)
+# Exercism Elm Track
+[![build status](https://travis-ci.org/exercism/elm.svg?branch=master)](https://travis-ci.org/exercism/elm)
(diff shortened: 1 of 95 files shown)


kytrinyx:
I just realized that all the package.json and elm-package.json files also have a reference to the repo in it.

tgecho:
The repo name is also duplicated through a bunch of exercise level metadata files. This evening I'll do a quick skim to remove any that aren't required and update the rest.

kytrinyx:
@tgecho Would you ping me if you need me to rebase this at some point?

tgecho:
Should be all set. Thanks!

kytrinyx:
Thanks @tgecho!

PR #150, opened 2017-07-05
Add "watch" command to all exercises

iamvery:
It's often nice when learning to be able to kick off a test command and glance at it while making changes. I really appreciated it about other Exercism tracks, and it seems like a nice addition to the Elm track 🙂

Thanks for such a wonderful resource for learning! ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Please keep the following in mind:
 - The recommended workflow when working on an exercise is to first create the implementation and test files, `ExerciseModuleName.elm` and `tests/Tests.elm`. You'll likely want to copy one of the existing exercise directories as a quick start.
   - Test the new exercise directly by running `npm test` from the exercise directory.
 
+  - Automatically run tests again when you save changes by running `npm run watch` from the exercise directory.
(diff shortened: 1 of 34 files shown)


tgecho:
Great point!

FWIW, you can always add stuff to npm commands with `--` (e.g. `npm test -- --watch`). This isn't terribly discoverable, though it could be documented.

Total bikeshed territory, but what about putting it under `start`? `npm start` is a few less keystrokes and does have a bit of precedent.

Neither of these are particularly serious objections. What do you think?

iamvery:
I don't think I have enough experience in the JS ecosystem to have terribly strong opinions. I have been using Yarn, so the command is just `yarn watch`. I'm not familiar with what sorts of things are typically kicked off with `npm start`. I went with `npm run watch` because that's the way the Ecmascript exercises are implemented. Figured there was some value in consistency. Maybe update them too if we agree?

Thanks for the tip about `--`! TIL 🌈 💫

tgecho:
Works for me! 👍

PR #152, opened 2017-07-05
Skip all but first test of each example

### LINKED ISSUE #151: Skipping tests?
Would there be any interest in skipping later tests in exercises? I'm thinking something similar to `xit` used in [the Ecmascript exercises](https://github.com/exercism/ecmascript/blob/master/exercises/accumulate/accumulate.spec.js). Syntactically it would be a little more heavy-weight in Elm, so I've opted to ask in an issue before making the change for pull request.

Also related, I noticed what seems to be an error in the elm-test README about this feature. See https://github.com/elm-community/elm-test/pull/186.

iamvery:
This allows learners to gradually approach exercise. However, unlike commented tests, the Elm compiler is still able to infer type information from the skipped tests.

Note: I made this change very mechanically, it's quite possible that the very first test is not the ideal starting point for learners. I'd really appreciate some review of the changes here before merge.

Thank you! ❤️ 

[Closes #151]

### DIFF
--- a/exercises/nucleotide-count/tests/Tests.elm
+++ b/exercises/nucleotide-count/tests/Tests.elm
@@ -10,11 +10,10 @@ tests =
     describe "NucleotideCount"
         [ test "the solution is for the correct version of the test" <|
             \() -> Expect.equal 2 version
-        , skip <|
-            test "empty dna strand has no nucleotides" <|
-                \() ->
-                    Expect.equal { a = 0, t = 0, c = 0, g = 0 }
-                        (nucleotideCounts "")
+        , test "empty dna strand has no nucleotides" <|
+            \() ->
+                Expect.equal { a = 0, t = 0, c = 0, g = 0 }
+                    (nucleotideCounts "")
         , skip <|
             test "repetitive sequence has only guanine" <|
                 \() ->
--- a/exercises/nucleotide-count/tests/Tests.elm
+++ b/exercises/nucleotide-count/tests/Tests.elm
@@ -10,16 +10,19 @@ tests =
     describe "NucleotideCount"
         [ test "the solution is for the correct version of the test" <|
             \() -> Expect.equal 2 version
-        , test "empty dna strand has no nucleotides" <|
-            \() ->
-                Expect.equal { a = 0, t = 0, c = 0, g = 0 }
-                    (nucleotideCounts "")
-        , test "repetitive sequence has only guanine" <|
-            \() ->
-                Expect.equal { a = 0, t = 0, c = 0, g = 8 }
-                    (nucleotideCounts "GGGGGGGG")
-        , test "counts all nucleotides" <|
-            \() ->
-                Expect.equal { a = 20, t = 21, c = 12, g = 17 }
-                    (nucleotideCounts "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC")
+        , skip <|
+            test "empty dna strand has no nucleotides" <|
+                \() ->
+                    Expect.equal { a = 0, t = 0, c = 0, g = 0 }
+                        (nucleotideCounts "")
+        , skip <|
+            test "repetitive sequence has only guanine" <|
+                \() ->
+                    Expect.equal { a = 0, t = 0, c = 0, g = 8 }
+                        (nucleotideCounts "GGGGGGGG")
+        , skip <|
+            test "counts all nucleotides" <|
+                \() ->
+                    Expect.equal { a = 20, t = 21, c = 12, g = 17 }
+                        (nucleotideCounts "AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC")
         ]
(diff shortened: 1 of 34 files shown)


tgecho:
So I thought of a problem a few hours ago, and Travis is backing me up with the [test failure](https://travis-ci.org/exercism/elm/builds/250512945?utm_source=github_status&utm_medium=notification): We still need to run all of the tests against the examples in CI. Unfortunately elm-test doesn't have a way to "force" running incomplete tests.

The best (but still bad) I can come up with offhand is to temporarily remove the skips (`sed 's/skip <\|//g'`) when we do the CI build. It's a bit gross, but I think optimizing for the end users is important enough here.

I have this commit prepared. Just doing another skim through the rest and I'll merge the lot.

exercises/nucleotide-count/tests/Tests.elm
```diff
@@ -10,16 +10,19 @@ tests =
     describe "NucleotideCount"
         [ test "the solution is for the correct version of the test" <|
             \() -> Expect.equal 2 version
```
tgecho: We should probably unskip the first actual test. The version check should automatically pass :)

exercises/run-length-encoding/tests/Tests.elm
```diff
@@ -10,28 +10,36 @@ tests =
...
-            \() ->
-                Expect.equal "WWWWWWWWWW"
-                    (decode "10W")
-        , test "encode unicode" <|
-            \() -> Expect.equal "⏰3⚽2⭐⏰" (encode "⏰⚽⚽⚽⭐⭐⏰")
-        , test "decode unicode" <|
-            \() -> Expect.equal "⏰⚽⚽⚽⭐⭐⏰" (decode "⏰3⚽2⭐⏰")
+        , skip <|
+            test "encode simple" <|
+                \() -> Expect.equal "2A3B4C" (encode "AABBBCCCC")
```
tgecho: Another first actual test 🔍

exercises/sublist/tests/Tests.elm
```diff
@@ -10,40 +10,58 @@ tests =
...
-            \() -> Expect.equal Superlist (sublist [ 1, 1, 1, 2 ] [ 1, 1, 2 ])
-        , test "superlist early in huge list" <|
-            \() -> Expect.equal Superlist (sublist (List.range 1 100000) [ 3, 4, 5 ])
-        , test "recurring values sublist" <|
-            \() -> Expect.equal Sublist (sublist [ 1, 2, 1, 2, 3 ] [ 1, 2, 3, 1, 2, 1, 2, 3, 2, 1 ])
-        , test "recurring values unequal" <|
-            \() -> Expect.equal Unequal (sublist [ 1, 2, 1, 2, 3 ] [ 1, 2, 3, 1, 2, 3, 2, 3, 2, 1 ])
+        , skip <|
+            test "empty equals empty" <|
+                \() -> Expect.equal Equal (sublist [] [])
```
tgecho: Another first actual test 🔎

exercises/robot-simulator/tests/Tests.elm
```diff
@@ -15,46 +15,51 @@ tests =
...
+                        \() -> Expect.equal South robot.bearing
                 ]
             )
         , describe "turn right"
             ((List.range 1 3)
                 |> List.scanl (\_ r -> turnRight r) defaultRobot
                 |> List.map .bearing
                 |> assertionList [ North, East, South, West ]
-                |> List.indexedMap (\i e -> test ("step " ++ toString i) (\() -> e))
+                |> List.indexedMap (\i e -> skip <| test ("step " ++ toString i) (\() -> e))
```
tgecho: I'm thinking this skip might be better at the describe level. elm-test won't let them forget it, but I think it'll be a bit more clear there. Also see the next describe block.

exercises/nucleotide-count/tests/Tests.elm
iamvery: I had noticed the version function on another exercise, but I don't think there was a test for it. What's the story with that?

exercises/robot-simulator/tests/Tests.elm
iamvery: That makes a ton of sense! Agree 👍

iamvery:
I didn't realize you could "force" any test runners to ignore skips! Do those options exist for exactly this sort of use case? Would it be a worthwhile addition to elm-test?

Is there anything else you would like me to do here, or are you taking over?

exercises/nucleotide-count/tests/Tests.elm
tgecho: Sometimes we change a test in a way that isn't backwards compatible. The version declaration in the exercise stub helps reviewers know what tests they were writing against. That's the idea at least... we haven't been perfectly consistent.

tgecho:
The only one I know of is nose (an old python test runner). Not too many use cases that I can think of, since IMO skips are usually temporary or a smell.

I have a commit prepped to do the skipping. If you don't have time to do the adjustments above I can take care of them in the next day or so to get this merged.

iamvery:
I believe I've added all the changes you requested. Thanks for your time and feedback :)

tgecho:
Thanks for your help! I think this is a really nice polishing up of the experience!

iamvery:
Thanks for the assists! ❤️

PR #153, opened 2017-07-06
Add version test to triangle exercise

iamvery:
This follows the pattern set by some of the other exercises. I noticed it was missing from the triangle exercise.

### DIFF
--- a/exercises/triangle/tests/Tests.elm
+++ b/exercises/triangle/tests/Tests.elm
@@ -2,7 +2,7 @@ module Tests exposing (..)
 
 import Test exposing (..)
 import Expect
-import Triangle exposing (triangleKind, Triangle(..))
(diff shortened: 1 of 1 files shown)


tgecho:
Good catch! Looks like you forgot to expose `Triangle.version` :)

iamvery:
Whoops! 😂  Fixed 💫

iamvery:
Curious, would it be preferable to access the function with the fully-qualified name `Triangle.version` rather than have it exposed to the `Tests` module? My thinking is that `version` is an awful generic name for a function. Perhaps I'm overthinking it here, but still interesting to me in general if that's a consideration...

tgecho:
A fair question. In general, I'm coming to prefer using qualified names more and more. This case is simple/static enough that I don't think it's worth any more effort :)

PR #163, opened 2017-10-10
Download elm-format on windows

mortenaa:
Here's what I had to do to make the build.sh script work on Windows (with git bash/mingw) in case you want to incorperate it


tgecho:
@anuragsoni Yep, you're right. I'll get that sorted and add a .gitattributes file to hopefully head this  off going forward.

tgecho:
Thanks all! 🎉

mortenaa:
Don't know what happened with the line endings there, thanks for sorting it out! :-)

PR #165, opened 2017-10-12
use isAlpha definition from the upcoming elm-lang core

anuragsoni:


exercises/isogram/Isogram.example.elm
```diff
@@ -5,16 +5,16 @@ import List
 import Char
 
 
-isLetter : Char -> Bool
-isLetter c =
-    Char.isHexDigit c && (not <| Char.isDigit c)
+isAlpha : Char -> Bool
+isAlpha char =
```
anuragsoni: same implementation as the future elm-core Char module

https://github.com/elm-lang/core/blob/master/src/Char.elm#L86

PR #168, opened 2017-10-26
Update Elm/Space-Age tests to use Float literals

nmbrgts:
The large Int literals in some of these tests result in errors (see https://github.com/elm-lang/elm-compiler/issues/1246) that can only be resolved by editing the test file. Changing the Int literals to Floats solves the issue. I think it would be best to fix  it here even though it does make the exercise more interesting.


tuxagon:
@nmbrgts @neganp I've re-opened this PR since it went stale during the maintainer hiatus

yurrriq:
LGTM, thanks!

PR #173, opened 2018-01-08
Update Phone-Number Tests.elm

Phippsaurus:
The format of _NANP_ phone numbers is described in the [README](https://github.com/exercism/elm/edit/master/exercises/phone-number/README.md) as 

`````
(NXX)-NXX-XXXX
`````

where `N` is a digit from 2 through 9, but all the tests for valid phone numbers test for `Just "1234567890"`.
That means the first digit of the _area code_, `N`, would be 1 there every time, for which a correct implementation should return `Nothing`.

My change would simply increment those 1s to 2, as i believe the intention is to test for valid numbers here.

Additionally I introduced two new test cases, expecting `Nothing`, when either digit `N` in the format is equal to 1.

### DIFF
--- a/exercises/phone-number/PhoneNumber.example.elm
+++ b/exercises/phone-number/PhoneNumber.example.elm
@@ -1,44 +1,60 @@
 module PhoneNumber exposing (..)
 
-import String
+import Regex exposing (Regex, regex, find, HowMany(..))
+import String exposing (concat, startsWith, length, slice)
+import List exposing (map, head)
+import Maybe exposing (andThen)
 
 
-nums =
-    [ '1', '2', '3', '4', '5', '6', '7', '8', '9', '0' ]
+matchesFormat : Regex -> String -> Bool
+matchesFormat regex string =
+    case find (AtMost 1) regex string |> head of
+        Just found ->
+            True
+
+        Nothing ->
+            False
 
 
 getNumber : String -> Maybe String
-getNumber text =
-    getValidNum (String.filter isDigit text)
+getNumber phoneNumber =
+    let
+        numberFormat =
+            regex "^\\D*(\\+?1[\\-\\s\\.]?)?((\\([2-9]\\d{2}\\))|([2-9]\\d{2}))[\\-\\s\\.]?[2-9]\\d{2}[\\-\\s\\.]?\\d{4}\\D*$"
 
+        numbers =
+            regex "\\d"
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rootulp** [comment]
@Phippsaurus thanks for looking into this! I'm not sure if you've seen the [problem-specifications] (https://github.com/exercism/) repo but it contains shared canonical data for exercism exercises. It seems like the elm phone-number tests didn't conform to the canonical data but you made them closer! I think there are a few tests you could add / edit to get closer if you're interested: https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json

**Comment by Phippsaurus** [comment]
Thanks for pointing me towards the canonical data, I added tests for those cases mentioned there which were not yet implemented.
I think the test cases should be okay now, I am just not sure if my example implementation is as straightforward as it should be.

**Comment by neganp** [comment]
Thanks for adding those, it looks like we'll have really thorough coverage now! :+1:

I like almost all of these tests, but I'm not sure about the leading and trailing character cases:

```elm
, skip <|
    test "valid with leading characters" <|
	\() -> Expect.equal (Just "2234567890") (getNumber "my number is 223 456 7890")
, skip <|
    test "valid with trailing characters" <|
	\() -> Expect.equal (Just "2234567890") (getNumber "223 456 7890 - bob")
```
If we read it strictly, these are outside the problem as stated:

> Your task is to clean up differently formatted telephone numbers by removing punctuation and the country code (1) if present.

I'm sure this is very pedantic of me, but I can imagine it feeling unfair for a learner to get through nearly all the tests only to be stumped by problems that they weren't told they'd have to solve. Does that make sense?

If you're looking for more feedback, I might move one of the `Nothing` result tests further up so that the learner can see both code paths being executed, but I can't see anything else to change.

The example implementation is very thorough! I didn't have any trouble following it; I think it can be committed as is.

Phippsaurus:
@neganp Thank you for your feedback, I interspersed some of the `Nothing` result tests with the valid inputs, so errors will be noticed sooner.

 As for the leading and trailing characters, I did not want to remove preexisting valid tests.

From a real world application point of view, I felt that it is more likely that a phone number has to be extracted from a paragraph of continuous text, as e.g. a form field would presumably impose narrower input restrictions, but on the other hand you are right that it feels unfair and does not add much learning benefit, so I removed those cases.

neganp:
Awesome, thanks!

Issue #174, opened 2018-01-15
rna-transcription: two error handling tests return Char


rootulp:
Two [tests](https://github.com/exercism/elm/blob/master/exercises/rna-transcription/tests/Tests.elm#L27
) on the rna-transcription exercise return type `Err Char` but it's not immediately clear why these `Char`s are chosen. The user may infer that the `Err` should return the first `Char` in the DNA strand but this isn't pointed out in the README. 

The [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/rna-transcription/canonical-data.json) doesn't appear to have these two error handling tests. I propose removing them or replacing them with a `Err String` where the String is something like `Invalid input`

petemcfarlane:
An alternative approach would be to update the README? Personally I didn't struggle to see that it returned the first invalid Char in the strand, I quite enjoyed the exercise as it is!

petemcfarlane:
A benefit of using a `Result` type is that you can capture unique information in the `Err`, otherwise you could just use a `Maybe`. If you always return `Err "Invalid input"` for any error, you may as well return a `Maybe.Nothing`. If you want to show why there was a failure you might do `Result.Err "'X' is not a valid nucleotide"` or `Result.Err "'Y' is not a valid nucleotide"`

rootulp:
I like your suggestion of `Result.Err "'X' is not a valid nucleotide"` where `X` is the first invalid nucleotide.

Issue #179, opened 2018-01-26
Hamming - return Result rather than Maybe


petemcfarlane:
Just a thought really, I thought it may be more interesting to use a `Result` return type, rather than a `Maybe` to handle the exception cases - it could be a good way to introduce people to the Result type too?

petemcfarlane:
PS happy to write/submit a PR if anyone thinks it's a good idea

rootulp:
The [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json) indicate these should be Errors but [currently](https://github.com/exercism/elm/blob/master/exercises/hamming/tests/Tests.elm#L48) they're `Nothing` b/c of the `Maybe` type. I'm in favor of replacing with a `Result` type that can have a `String` error similar to the canonical data.

petemcfarlane:
I've added a PR, but since opening this issue, I'm not sure there's a whole load of benefit in using a Result when the error message is always the same? Maybe that's just me learning about using `Result` types. What do other people think? Is a `Result` more useful when there's more than one way something can fail? Or does `Result` still imply more than a `Maybe` with it's name, as it's the result of a function/calculation? What would you likely use in a production app?

neganp:
I would agree that a `Result` type is better for this exercise because there are several ways the function could fail, and we only *happen* to see one. There could, for instance, be an invalid nucleotide. It might be worth looking in to getting such an example into the problem spec.

tuxagon:
@neganp, that's a fair point. In addition to changing the spec, it would also be better to not use a string and have a union type along the lines of `type Error = Invalid | DifferentLengths`. That would be more idiomatic elm because you can rely on types more heavily instead of just the `String` type. I say it is more idiomatic because then the caller could handle the different errors instead of matching on strings, which is very error prone.

neganp:
That's a good thought! :+1:

petemcfarlane:
Ok, so the only errors we have are DifferentLengths right now, do you think another error we could test for and add to the error union type would be `InvalidNucleotide 'X'` for example?

tuxagon:
@petemcfarlane I think that makes sense; however, I didn't see anything in the README that details what "invalid" means. Adding it would feel more code complete, but what is considered invalid would probably need detailed out in the README. For this specific issue, it might be too much. Do you agree @neganp?

neganp:
I think, based on this thread, the more complicated fix would look something like this:
 
[...]

PR #180, opened 2018-01-29
Improve the error message in the Result

petemcfarlane:
Instead of just returning a Char, return a string including the first erroneous Char. Reference #174


neganp:
Thanks for submitting this! I would agree that a more explicit error message would be helpful.

As discussed over in #179, what do you think of returning a union type instead?

PR #181, opened 2018-01-29
Return a Result instead of Maybe

petemcfarlane:
Reference #179


neganp:
Thanks, this looks really good! Sorry it took so long to get to!

PR #187, opened 2018-02-06
Inline the exercise README insert

kytrinyx:
Now that we're operating with a track-wide template for generating exercise
READMEs we no longer need a separate markdown document to use for hints
and instructions that apply to all the exercises in a track. These instructions
can be defined directly in the template.

The first commit regenerates the exercise READMEs to pick up tweaks to problem descriptions, which have been updated with various formatting changes, whitespace changes, and clarifications to the description.

That gives us a clean diff when inlining the template and regenerating.

Ref: https://github.com/exercism/meta/issues/94


tuxagon:
@neganp Are we waiting for anything else on this one?

neganp:
Because I'm not super familiar with how track generation works, I was either going to stand up a dev environment and run through the whole process or wait for someone more knowledgeable to review this change. I don't really have time to do the former, and it looks like we might be a while waiting for the latter. If this is blocking work, I think it would be fine to merge it.

tuxagon:
I've played around with the generation and it's actually pretty easy. I'm not sure what the scope of the dev environment you want to set up is, but there are only a few steps to recreate the README generation that was done on this PR. I understand it a lot better now.

1. Download `configlet` from https://github.com/exercism/configlet/releases
2. Clone `exercism/elm` and `exercism/problem-specifications` at the same level (like `$HOME/elm` and `$HOME/problem-specifications`)
3. Run `configlet generate $HOME/elm`

That should change everything according to this part of configlet's documentation: https://github.com/exercism/configlet#generate

I could merge it you want or let you wait since you are marked as the reviewer. If you already knew all of that, I apologize for being verbose; hopefully, it is helpful to someone, though.

There are still a few missing links for me, like `config` vs `.meta` and where inlining vs using `docs` is, but I think some of that is in flux, currently.

churchie317:
I think I would be in favor of merging.

As @tuxagon pointed out track generation can seem complex on its face, but once you get to playing around with it it's actually pretty straight forward.

RE:

> There are still a few missing links for me, like config vs .meta and where inlining vs using docs

My understanding is that the `.meta` directories are only used in the cases when we want to override the default READMEs. Something--and please correct me if I'm wrong--we don't currently need to worry about.

kytrinyx:
> My understanding is that the .meta directories are only used in the cases when we want to override the default READMEs. Something--and please correct me if I'm wrong--we don't currently need to worry about.

Yepp—at the moment, `.meta` is only necessary if you're overriding the defaults. 👍

churchie317:
@neganp 

> I was either going to stand up a dev environment...

I wrote a little script that should prepare a basic dev environment for you [here](https://gist.github.com/churchie317/e4cb78ab6e86c20bf9cbb6024b18fd3c). It will create a directory called `exercism-dev` and install the necessary dependencies, including `configlet`. Assuming the environment setup is successful, you can generate, lint, format, etc. the track by running:

1. `cd exercism-dev` followed by, for example;
2. `elm/bin/configlet generate elm/ -p problem-specifications/` to generate exercises.

Hopefully that helps!

PR #189, opened 2018-02-06
Skip all tests but first in all-your-base

bobwhitelock:
To avoid overwhelming the user with many test failures as soon as they start the exercise. This is also consistent with every other Elm exercise, which I have confirmed using `git grep -L ' skip ' -- 'exercises/**/tests/*.elm'`.


neganp:
Looks good!

Issue #194, opened 2018-02-12
Improve Consistency Among Exercises


tuxagon:
I decided to take a stroll through the current exercises and found some inconsistencies when it comes to `package.json`.

First, there is a root `package.json`
```json
{
  "scripts": {
    "test": "elm-test"
  },
  "dependencies": {
    "elm": "0.18.0",
    "elm-test": "0.18.3"
  }
}
```

In *_most_ of the exercises
```json
{
    "description": "Exercism/Elm",
    "repository": "https://github.com/exercism/elm.git",
    "license": "MIT",
    "scripts": {
        "postinstall": "elm-package install -y",
        "watch": "elm-test --watch",
        "test": "elm-test"
    },
    "dependencies": {
        "elm": "0.18.0",
        "elm-test": "0.18.3"
    }
}
```

Also, as of this issue creation, none of the exercises have `package-lock.json`, which is fairly new (added in npm version `^5.x.x`).

I'm not sure what the decision behind each exercise having their own `package.json` is, but it could be removed altogether and kept only at the root level. This would help any updates be more consistent and it would mirror the same approach https://github.com/exercism/javascript does by just having a root `package.json`.

*: Non-existent in `all-your-base`

churchie317:
So the proposal is to drop the `package.json`s for each exercise, define one at the root _and_ define a single `package-lock.json` at the root as well, correct? If so, I can probably take care of this.

tuxagon:
@churchie317 Unless there are objections, that is what makes sense to me. The `package.json` files in each exercise feel superfluous. The `package-lock.json` file will just get created automatically for you if you clone and then `npm install` at the root of the repo.

If you are able to tackle this, let me know if you actually encounter issues. This issue was created to start the conversation and then have a clear direction moving forward so that it can be part of the PR checklist.

churchie317:
Okay, so looking into it I suspect there's a reason each exercise includes its own `package.json` and it's related to the exercise `README` instructions. I suspect the instructions were assembled at a time before it was idiomatic to install Elm globally. Specifically, the `README` instructions are to run:

1. `npm install`. Looking at the `package.json`, we can see this means Elm is being installed for each exercise. I'd be interested to hear some thoughts, but this feels heavy-handed to me.
2. `npm test` which again is merely a wrapper for running `elm-test`.

The only nice-to-have users would lose is the test watcher, but we can tell them to simply run `elm-test --watch`. 

[...]

Issue #182, opened 2018-01-29
Include instructions for globally installing dependencies


neganp:
As written, the instructions for running the tests have NPM installing `elm` and `elm-test` in every exercise directory. This is fine, but could be avoided by installing them globally by running

    npm install -g elm elm-test

and setting up the executables once, instead of running 

    npm install

and waiting ~10 seconds in each exercise directory.

I'm by no means a Node expert, but is there a compelling reason to not do this?

tuxagon:
`npm install` does become superfluous if you've installed globally, meaning you wouldn't have to run it again even if you changed directories. The only thing I can think of is if someone wants to keep things either 1) isolated or 2) they have a different version installed for something else and they don't want to overwrite it globally.

Are you referring to just changing the `README` in each project? I just want to be clear.

neganp:
Clarity is good! 👍 

I'm talking about changing the [installation instructions](http://exercism.io/languages/elm/installation) and maybe the README template. I think installing the dependencies once is more palatable: having one copy of the deps feels tidier, and being able to run the code without the install step makes starting a new exercise a little bit quicker.

tuxagon:
Ah, I see. I would say installing globally for the installation instructions is a good idea, then. Reason for which is that elm's installation instructions even say that. If it affects someone, the chances are good that they will know and take any necessary actions to accommodate their special circumstances.

https://www.npmjs.com/package/elm
https://www.npmjs.com/package/elm-test

churchie317:
> I'm by no means a Node expert, but is there a compelling reason to not do this?

The only reason this might be discouraged in the Node community is because global dependencies are generally reserved for local tooling, and the community really encourages an application's dependencies to be listed in the `package.json`. Obviously, requiring something to be globally installed--i.e. absent from the dependency manifest--makes downloading an application's dependencies a little trickier than a simple `npm i` and can lead to scenarios where an application works in one environment but not others.

That said, I agree that changing the installation instructions and updating the README template is a good idea. We definitely want to encourage the community to write Elm in an idiomatic way and the community has unequivocally sided on `npm install -g elm-test`.

tuxagon:
#198 closes this

PR #170, opened 2017-12-01
Implemented twelve-days exercise

tuxagon:
I went ahead and implemented the `twelve-days` exercise. Let me know what I need to do to get it accepted.

PR #198, opened 2018-02-23
Consistency

tuxagon:
This takes care of #194 and #182. The gist of the changes include
* Removing exercise specific `package.json` files
* ~Updated `bin/build.sh` to use `elm-test` explicitly~
* Changed installation instructions to favor a global install
* Updated template and regenerated all READMEs

Outstanding questions include
* What are your thoughts on the new installation instructions since they needed to be updated to reflect `elm-test` too.
* Do you see anything I missed?


neganp:
Nice! :clap: 

Nothing jumps out at me after a brief review.

If I had to nitpick, I'd say this PR might be improved by splitting it up (e.g: update docs, update package.json, then update the exercises), but I don't think it's a problem in this case.

Thanks for taking care of all these. :+1:

tuxagon:
@neganp I thought about that, but I didn't think it would be a good idea to have it be in a non-consistent state while separating the logical pieces. I instead went for very granular focused commits instead.

Thoughts?

neganp:
I think that makes sense. :slightly_smiling_face: 

Would you be okay leaving this open over the weekend to give folks a chance to comment and merging it on Monday if nothing comes up?

tuxagon:
I'm fine with that because I also want to update #170 to follow this. Do you want to merge it or me?

neganp:
Whoever gets to it first :+1:

tuxagon:
The race is on 😉

PR #203, opened 2018-02-26
New Exercise: Acronym exercise

tuxagon:
I went ahead an implemented the acronym exercise. It is also already updated to include the changes made in #198

### DIFF
--- a/config.json
+++ b/config.json
@@ -333,6 +333,16 @@
       ]
     },
     {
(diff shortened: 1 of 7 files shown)


neganp:
Nice! This looks real good :100:

If there are no objections, I'll leave this for a day or two in case anyone wants to comment, and then I think it's ready to merge.

Thanks for your contribution :+1:

PR #205, opened 2018-03-03
Implemented transpose exercise

tuxagon:
Another one down. This exercise was actually really fun to do because of the jaggedness.

### DIFF
--- a/config.json
+++ b/config.json
@@ -392,6 +392,15 @@
       ]
     },
     {
+      "uuid": "66fef1df-aa12-40eb-b572-7ea9b6a920ed",
+      "slug": "transpose",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 5,
+      "topics": [
+      ]
+    },
+    {
       "uuid": "1F9FE5BC-8213-44FD-B7D1-5D4CC7F3A475",
       "slug": "binary-search",
       "core": false,
--- a/exercises/transpose/README.md
+++ b/exercises/transpose/README.md
@@ -1,0 +1,95 @@
+# Transpose
+
+Given an input text output it transposed.
+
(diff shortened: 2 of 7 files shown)


neganp:
Nice work! :+1: This is a neat problem, thanks for writing it up.

It looks good except for a couple of things:

1. Could you add a stub implementation of the `transpose` function in the template (using, for example, [`crash`](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Debug#crash) with a message as the function body), as well as the imports for String and List? This is to be up-front about the function signature and data-types the learner is expected to use, so that (for example) they don't start solving it with Arrays or string bashing and then have to change their solution.

2. Could you [`skip`](http://package.elm-lang.org/packages/elm-community/elm-test/4.2.0/Test#skip) all the tests except the first one?

These things are sort of documented in the [exercism docs](https://github.com/exercism/docs/blob/0c5d53ad2a4db734a6e95d6eacfe0382c627d487/you-can-help/review-pull-requests.md#consider-the-user-experience), but this track isn't great about doing them. :slightly_frowning_face:

tuxagon:
@neganp I forgot to add the `skip` back in! Thanks for noticing. As for the first one, do we want this to be the first problem that has a stub? I'm completely fine with it and I'm actually a bigger fan of having a stub; it's just not present in the other exercises. 

What I could do is potentially stub this one and then create a PR to stub the rest of them. Thoughts?

neganp:
I think that sounds like a very sensible plan. :+1: I've opened an issue (#206) to track this.

tuxagon:
Alright, I'll make that change, probably, sometime tonight.

neganp:
I think this has waited long enough. Thanks again for the new exercise! :+1:

PR #208, opened 2018-03-30
Ran configlet fmt

### LINKED ISSUE #200: Run fmt with configlet
I noticed that running `configlet fmt` will make changes to the `config.json` and `config/maintainers.json` files. It would probably be a good idea to run that and get all the files up to expectations according to `configlet`. Ultimately, the changes are just re-organization of fields, but it would allow the command to be run in an isolated setting without affecting everything.

tuxagon:
Resolves #200 

This is really just from running `configlet fmt`

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,419 +1,344 @@
 {
-  "language": "Elm",
   "active": true,
   "exercises": [
     {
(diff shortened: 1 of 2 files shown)


tuxagon:
Since this is just from running `configlet fmt` and it has been a few days, I'm going to merge it.

PR #209, opened 2018-05-02
New script for basic new exercise setup

tuxagon:
I added a new script to the `bin` directory for generating a basic new exercise setup. Ultimately, this could be expanded upon later, especially if we get to test generation (although it should probably not be bash at that point). 

It'd be nice if the README's contributing section can be looked over for typos or if something is missing or needs better explanation. There might also be dead information. 

Hope this helps lower the barrier to entry for new contributors for implementing exercises, which the elm track desperately needs a lot of.


neganp:
This looks fine. I'm a little hesitant to add a bunch of code for an activity that might be carried out just a few dozen times, but I can see how it would make it easier for a new contributor to get the hang of how to write an Exercism case.

tuxagon:
@neganp, that's fair. It's a little something I cooked up and wanted to share the process to help alleviate the pain of knowing where to start to add a new exercise. Elm is really behind in exercises and I thought this might help close that gap.

tuxagon:
@neganp do you have other comments? good to merge? hold off?

PR #210, opened 2018-05-02
Update TESTS.md

rofrol:


tuxagon:
Hi @rofrol, thanks for the contribution! It was a good catch; documentation like that falls through the cracks sometimes.

I think everything looks good, especially since it's just a README change. I'm going to go ahead an merge.

Thanks again!

PR #211, opened 2018-05-03
Update TESTS.md - install needed

rofrol:


tuxagon:
Hello again, @rofrol. You may not know this, but running `elm-package install` is unnecessary because `elm-test` will pick up the slack for you when you run the command the first time.

By just relying on `elm-test` here, we can reduce the extra complexity for setting up for an exercise.

You can confirm this with 
```bash
exercism fetch elm hello-world
cd ~/exercism/elm/hello-world
elm-test
```

rofrol:
Somehow `elm test` didn't want to run without `install` first, but I cannot reproduce it now. Closing,.

PR #213, opened 2018-05-08
run-length-encoding: Make test cases match canonical test spec

mrattner:
The run-length-encoding exercise benefits from having a property-based (fuzz) test to make sure it works on a wide range of inputs. Previously the test code seemed to be using the elm-check package for this purpose, but the RunLengthEncodingPropertyChecks file is still hanging around even though it's not currently being used.

I updated the tests file to use elm-test's fuzzing capability, which seems to do what RunLengthEncodingPropertyChecks intended.

**Edit:** I did run elm-format on the file before committing, so I'm puzzled at the build error. I'm happy to make any further changes to conform with whatever the auto build requires, but would need a tip on what might be wrong :)


tuxagon:
Hello @mrattner, thank you for your contribution. That file, indeed, was just left as an artifact from a bygone age. There are 2 suggestions I can make
1. Use `bin/install-elm-format`, which will install the correct `elm-format` into the `bin/` directory. Whatever version that script installs is considered the canonical version for the elm track.
2. Since we are re-visiting the tests for the `run-length-encoding` exercise, I noticed the tests differ from the canonical tests that should be shared across tracks. If you are unfamiliar with what I am referring to, you can find them [here](https://github.com/exercism/problem-specifications/blob/master/exercises/run-length-encoding/canonical-data.json)

Let me know if there is any questions 😄

mrattner:
Does the canonical test spec mean that each language's implementation of that exercise shouldn't have other tests (i.e., the fuzz test, and the unicode emoji tests) in addition to the canonical ones? Anyway, I edited the test file to at least make it a superset of the canonical tests.

mrattner:
I see that the example implementation seems to have failed a fuzz test (given the string "+"). I could update the example to an implementation that passes the fuzz test (well, passes with that seed, and the seeds I've tried so far :wink:) Or I can remove/alter the fuzz test.

tuxagon:
@mrattner, I think you should just remove the fuzzing because it doesn't really add to the exercise in a meaningful way, and it looks like any numeric values in the input are not handled nicely based on how the README describes it. The README specifically says 

> For simplicity, you can assume that the unencoded string will only contain the letters A through Z

Based on the README, I think we should remove any tests that do not deal with characters between A and Z, like the emoji. They are tests that add an extra complication, but ultimately don't add much to the solution and diverge from the generated README.

mrattner:
Makes sense. Thanks for the input!

tuxagon:
Once again, I would like to thank you for your contribution, @mrattner. Cheers!

PR #215, opened 2018-05-12
phone-number: Remove extra function to match canonical tests

mrattner:
I noticed when starting the PhoneNumber exercise that the prettyPrint function is not part of the problem specification (I guess it used to be, in a previous version of the exercise). I thought that might be confusing since there is no mention of prettyPrint in the exercise instructions.

This commit removes prettyPrint and the tests that referred to it, as well as renaming & reordering the tests for PhoneNumber to match the current [canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json).

### DIFF
--- a/exercises/phone-number/PhoneNumber.elm
+++ b/exercises/phone-number/PhoneNumber.elm
@@ -1,11 +1,6 @@
-module PhoneNumber exposing (getNumber, prettyPrint)
+module PhoneNumber exposing (getNumber)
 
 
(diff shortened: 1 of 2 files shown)


tuxagon:
Hello again, @mrattner. Thank you for another contribution! On my end, this all seems fine. Nice catch with the mysterious extra function. I'm going to go ahead and merge because there isn't really a reason to wait. Thanks, again!

PR #218, opened 2018-06-10
Reformat config files

kytrinyx:
This runs the configlet fmt command, which normalizes the
contents and ordering of keys in the track config and
maintainers config.

This will let us script changes to the config files without
having unnecessary noise in the diffs when submitting pull
requests.

See exercism/meta#95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,352 +1,352 @@
 {
+  "language": "Elm",
   "active": true,
+  "blurb": "",
   "exercises": [
     {
-      "core": false,
-      "difficulty": 1,
       "slug": "hello-world",
-      "topics": null,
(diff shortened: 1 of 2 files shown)


tuxagon:
@kytrinyx this looks good. Was this just an explicit ordering of the keys vs the default alphabetical ordering that JSON marshaling does in go?

kytrinyx:
> Was this just an explicit ordering of the keys vs the default alphabetical ordering that JSON marshaling does in go?

Yeah, mostly explicit vs alphabetical, but also a couple of tweaks (add a newline to the end of the file for POSIX compliance, and distinguish between `null` topics and an empty list, where `null` indicates that it's not been processed yet, whereas an empty list indicates that there are no interesting, relevant topics).

PR #220, opened 2018-06-13
Update for v2

tuxagon:
This should complete #219 for the v2 release

There is a lot going on with this PR. To help push this along, I leveraged heavily from other tracks like fsharp, so there are definitely things that could be more accurate in some way. I'm confident that topics and difficulty are going to change at some point in the future. For the v2 launch, this can be the baseline.

@neganp, you're the one that has helped the most, so if you have the time, a review of whatever capacity you can would be helpful. Anyone else is also welcome to review.

😅

Edit: Reviewing `config.json` will be easiest at d17958c and before. The re-order really screws with the diff.

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,7 +1,9 @@
-Elm's goal is to be the best of functional programming in the browser.
+Elm is a pure, functional language that compiles to javascript.
 
-One of the primary goals of Elm is to completely eliminate runtime exceptions before they reach production by combining an aggressively helpful compiler with solid architectural principles.
+It was designed by Evan Czaplicki as part of writing his thesis on [Concurrent FRP for Functional GUIs](http://elm-lang.org/papers/concurrent-frp.pdf) and evolved into a language to help make making user interfaces better.
 
-Elm was designed by Evan Czaplicki as part of writing his thesis on [Concurrent FRP for Functional GUIs](http://elm-lang.org/papers/concurrent-frp.pdf).
+Elm's goal is to be the simplest and best language to use to make web-based user interfaces.
+
+One of the primary goals of Elm is to completely elminate runtime exceptions before they reach production by combining an aggressively helpful compiler with solid architectural principles.
 
 You can learn more about Elm at the [official website](http://elm-lang.org/).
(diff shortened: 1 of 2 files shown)


docs/ABOUT.md
```diff
@@ -1,7 +1,9 @@
-Elm's goal is to be the best of functional programming in the browser.
+Elm is a pure, functional language that compiles to javascript.
 
-One of the primary goals of Elm is to completely eliminate runtime exceptions before they reach production by combining an aggressively helpful compiler with solid architectural principles.
+It was designed by Evan Czaplicki as part of writing his thesis on [Concurrent FRP for Functional GUIs](http://elm-lang.org/papers/concurrent-frp.pdf) and evolved into a language to help make making user interfaces better.
 
-Elm was designed by Evan Czaplicki as part of writing his thesis on [Concurrent FRP for Functional GUIs](http://elm-lang.org/papers/concurrent-frp.pdf).
+Elm's goal is to be the simplest and best language to use to make web-based user interfaces.
```
neganp: I would suggest changing this to 

> Elm's goal is to be the simplest and best language for making web-based user interfaces.

to avoid repeating "to" twice in such close proximity.

docs/ABOUT.md
```diff
@@ -1,7 +1,9 @@
...
+
+One of the primary goals of Elm is to completely elminate runtime exceptions before they reach production by combining an aggressively helpful compiler with solid architectural principles.
```
neganp: Typo: s/elminate/eliminate

neganp: It's a little odd to have "Elm's goal" above and then "Elm's goals" here. Some learners (who are perhaps new to coding or don't come from a software engineering background) might also be unfamiliar with the colloquial use of "development" and "production" as environments.

I might change this sentence to 

> Elm aims to completely eliminate runtime exceptions by combining an aggressively helpful compiler with solid architectural principles.

docs/ABOUT.md
tuxagon: Good call!

tuxagon:
If there are no further change requests, I plan on merging this at the end of the week, then.

Issue #227, opened 2018-08-19
Remove obsolete version tracking assertions in exercises


kytrinyx:
Some tracks have added assertions to the exercise test suites that ensure that the solution has a hard-coded version in it.
In the old version of the site, this was useful, as it let commenters see what version of the test suite the code had been written against, and they wouldn't accidentally tell people that their code was wrong, when really the world had just moved on since it was submitted.

If this track does not have any assertions that track versions in the exercise tests, please close this issue.

If this track _does_ have this bookkeeping code, then please remove it from all the exercises.

See https://github.com/exercism/exercism/issues/4266 for the full explanation of this change.

tuxagon:
For tracking purposes, not all of the elm exercises have this version bookkeeping. Here is a list of the found ones for quick reference when this gets completed
* https://github.com/exercism/elm/blob/c6f0a5251755c512f54752a85f5449a95efcefef/exercises/nucleotide-count/tests/Tests.elm#L11
* https://github.com/exercism/elm/blob/c6f0a5251755c512f54752a85f5449a95efcefef/exercises/run-length-encoding/tests/Tests.elm#L11
* https://github.com/exercism/elm/blob/c6f0a5251755c512f54752a85f5449a95efcefef/exercises/sublist/tests/Tests.elm#L11
* https://github.com/exercism/elm/blob/c6f0a5251755c512f54752a85f5449a95efcefef/exercises/triangle/tests/Tests.elm#L11

mpizenberg:
Seems like this has been dealt with so I'm closing the issue.

PR #231, opened 2018-10-05
Removed depricated version tests

ulve:
This should solve #227

No more expect version


ulve:
Yes! Will do!
fre 5 okt. 2018 kl. 11:27 skrev Nathan Parsons <notifications@github.com>:

PR #235, opened 2018-10-18
elm test should be elm-test

CBedzz:
change docs to correct to change ```elm test --watch``` to ```elm-test --watch```


tuxagon:
@CBedzz Sorry for not responding sooner. Nice catch! Looks good to me, so I'll go ahead and merge this.

PR #237, opened 2018-10-22
Update docs to install correct elm-test package

estiens:
We need to install elm-test@beta with elm version 0.19.0 or the tests won't run.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -7,7 +7,7 @@ it from [the official site](https://nodejs.org/). Once you have Node.js up and
 running, you can install elm and elm-test using 
 
 ```bash
-npm install -g elm elm-test
+npm install -g elm elm-test@beta
 ```
 
 which will install elm and place any tooling on `$PATH`.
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -7,7 +7,7 @@ it from [the official site](https://nodejs.org/). Once you have Node.js up and
 running, you can install elm and elm-test using 
 
 ```bash
-npm install -g elm elm-test
+npm install -g elm elm-test@beta
```
tuxagon: Actually, since we are using elm 0.19, we should be more specific and use `elm0.19.0` There are 4 different tags you can see [here](https://www.npmjs.com/package/elm-test/v/0.19.0-beta11). This will help us later when upgrading to the next version.

estiens: :+1:

estiens:
@tuxagon - do you want to target a specific version of elm in the installation instructions each time as well?

That way if there are breaking changes in either elm or elm-test they move in lockstop?

tuxagon:
@estiens That's a good idea. It's a little more verbose, but it removes the ambiguity of which version to use. The tag is `elm0.19.0` 

https://www.npmjs.com/package/elm

estiens:
@tuxagon sorry for delay in circling back to this, docs updated now

tuxagon:
Nice! Thanks @estiens! I'll go ahead and merge now.

PR #244, opened 2019-02-26
Update the type of `School`

leojpod:
The School definition should use the Grade type


leojpod:
I think it makes more sense this way since all the other functions are using `Grade` in their type signature.

tuxagon:
I like it, @leojpod! I'm going to go ahead and merge it.

leojpod:
Awesome :)

PR #247, opened 2019-03-13
updates test song text to match read me's song text

rebelwarrior:
changes a comma to a colon in `my true love gave to me,` to `my true love gave to me:`.


rebelwarrior:
DO NOT accept before reading this:
This update will make the example code fail since it relies on "," 

I can supply a new answer but mine doesn't have the circular way the example works.

rebelwarrior:
Needs this one: https://github.com/exercism/elm/pull/248

PR #249, opened 2019-03-13
Updates Twelve Days exercise test and example to match ReadMe

rebelwarrior:
ReadMe uses a ":" while the tests and example used a "," this corrects the test and example to match the read me.

### DIFF
--- a/exercises/twelve-days/TwelveDays.example.elm
+++ b/exercises/twelve-days/TwelveDays.example.elm
@@ -21,10 +21,10 @@ lyric day =
 
         ending =
             if day == 1 then
-                concat [ ", ", gift day ]
+                gift day
 
             else
-                verseEnd day
+                verseEnd day day
     in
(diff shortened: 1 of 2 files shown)


tuxagon:
Hello again @rebelwarrior 😄 I like the changes; however, there is one more thing I need to mention. In case you don't know, there is a repo that represents the canonical data for tests. We can change them downstream fine, but it is always a good idea to also consider creating a PR so that it can trickle down to other language tracks. 

Some tracks actually have a code generator to generate the tests based on the canonical test data. The elm track currently doesn't have this, but it is on my todo list at some point. So without the upstream change, the comma is going to come back _if_ we ever get a test generator.

https://github.com/exercism/problem-specifications/blob/master/exercises/twelve-days/canonical-data.json

It's definitely not an initially obvious thing about exercism, so I thought I'd share.

PR #250, opened 2019-04-15
space-age: add more tests, use problem specification data

michaelglass:
1. this PR uses floating point comparison, so that we can bring in values from problem specification instead of rounded values

2. without more tests, you can hard code one age per planet

### DIFF
--- a/exercises/space-age/tests/Tests.elm
+++ b/exercises/space-age/tests/Tests.elm
@@ -1,6 +1,6 @@
 module Tests exposing (tests)
 
-import Expect
+import Expect exposing (FloatingPointTolerance(..))
 import SpaceAge exposing (Planet(..), ageOn)
 import Test exposing (..)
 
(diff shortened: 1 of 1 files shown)


tuxagon:
Hi @michaelglass, thanks for the contribution! In case you are unfamiliar with exercism's setup, there are actually canonical tests that should be fixed upstream to solve these problems as well 
https://github.com/exercism/problem-specifications/blob/master/exercises/space-age/canonical-data.json

We can add the tests here first, but I highly encourage and recommend making a PR upstream as well.

As for this particular PR, elm-format will need to be ran using `bin/install-elm-format && bin/elm-format` to address the build server issue.

michaelglass:
@tuxagon made a PR upstream

tuxagon:
@michaelglass this looks good! I'm going to go ahead and merge now.

michaelglass:
You might have to revert this. Apparently the discussion went the other way on the problem specification PR

tuxagon:
@michaelglass thanks for following up. I'll wait and see what happens there first. I, for sure, didn't expect that outcome.

PR #252, opened 2019-04-19
binary-search: return Nothing instead of -1

michaelglass:
I don't know if this warrants a upstream PR.

For the binary search algorithm, we have students returning `-1`. This leads to bad habits and can be harder to implement. This PR replaces `-1` with `Nothing`

### DIFF
--- a/exercises/binary-search/BinarySearch.elm
+++ b/exercises/binary-search/BinarySearch.elm
@@ -3,6 +3,6 @@ module BinarySearch exposing (find)
 import Array exposing (Array)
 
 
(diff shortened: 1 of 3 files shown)


tuxagon:
I think you could argue that this doesn't need to be upstream because the return value is going to be heavily dependent on the language and idioms of that community.

michaelglass:
thank you!

tuxagon:
This looks good to me, I'll go ahead and merge. Thanks for another contribution, @michaelglass!

Issue #255, opened 2019-06-25
Grains exercise: Add tests for second condition.


MatsLanGoH:
## Problem
(**This probably makes for an easy first issue.** )

The exercise requires a solution to pass the following two conditions:

	Write code that shows:

	- [x] how many grains were on a given square, and
	- [ ] the total number of grains on the chessboard

However, the included test code only checks for the first part of the solution.

## Suggestion

Add the following bits to the `grains` exercise.

- [ ] Add stub code for the `total` function.
- [ ] Add [test code](https://github.com/exercism/elm/blob/2f1ca175263bed54d4bd73a7e193df996df14003/exercises/grains/tests/Tests.elm
) for the "total number of grains" part of the solution

  - for example:

  ```Elm
        [ describe "total"
            [ test "of 1" <|
                \() -> Expect.equal (Just 1) (total 1)
            , 
                test "of 2" <|
                    \() -> Expect.equal (Just 3) (total 2)
  ```

tuxagon:
@MatsLanGoH Thanks for the issue. I've updated the labels.

MatsLanGoH:
I went ahead and made some additions to the exercise :)
Could someone review #256 in their spare time?

mpizenberg:
Closing this since all concerns are discussed in associated PR

PR #256, opened 2019-07-03
Add tests for Grains exercise

MatsLanGoH:
This addresses the changes proposed in #255:
- adds tests for "total number of grains" part
- adds stub code for `total` function
- adds example implementation for `total` function

### DIFF
--- a/exercises/grains/Grains.elm
+++ b/exercises/grains/Grains.elm
@@ -1,6 +1,11 @@
-module Grains exposing (square)
+module Grains exposing (square, total)
 
 
 square : Int -> Maybe Int
 square n =
+    Debug.todo "Please implement this function"
+
+
+total : Int -> Maybe Int
+total n =
     Debug.todo "Please implement this function"
--- a/exercises/grains/Grains.example.elm
+++ b/exercises/grains/Grains.example.elm
@@ -1,4 +1,4 @@
-module Grains exposing (square)
+module Grains exposing (square, total)
 
 
 square : Int -> Maybe Int
@@ -8,3 +8,12 @@ square n =
 
     else
         Just <| 2 ^ (n - 1)
+
+
+total : Int -> Maybe Int
+total n =
+    if n < 1 then
+        Nothing
+
+    else
+        Just <| 2 ^ n - 1
--- a/exercises/grains/tests/Tests.elm
+++ b/exercises/grains/tests/Tests.elm
@@ -1,7 +1,7 @@
 module Tests exposing (tests)
 
 import Expect
-import Grains exposing (square)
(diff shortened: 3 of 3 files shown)


tuxagon:
@MatsLanGoH Thanks for the contribution! In case you're unaware, each exercism problem has canonical tests. As I reminded myself about this problem, you're right about needing the `total` function, but it looks like the `total` function only returns the total number of grains on the chessboard.

For reference, [here](https://github.com/exercism/problem-specifications/blob/master/exercises/grains/canonical-data.json#L98) is the canonical test for `total`

Also, if you're wondering, [here](https://github.com/exercism/haskell/blob/master/exercises/grains/test/Tests.hs#L41) is a similar test implementation for haskell

It looks like there might be some confusion on what the goal of this method is and I hope these links help provide some context.

Let me know if you have questions!

MatsLanGoH:
@tuxagon 
Oh, you're right! It seems I read more into the `total` function than there was to it.

Do you think it's still testable in the context of [the issue](https://github.com/elm-lang/elm-compiler/issues/1246) reported in the test case comments? 

If yes, I would just go on to change both `total` function and test implementation to match the Haskell version. Otherwise I'll just quietly close this PR and go looking for other opportunities to contribute ;)

tuxagon:
@MatsLanGoH Well, that is unfortunate. This is something to keep an eye on but looks like we will need to wait until elm can handle those higher numbers.

mpizenberg:
Hi @MatsLanGoH and @tuxagon,

Elm compiles to JavaScript and so this is a JS issue. In fact, JS only has one number type, which are 64bit floats. Since it keeps bits for the mantissa (cf https://en.wikipedia.org/wiki/Floating-point_arithmetic), it can only represent integers safely until 2^53 - 1 (cf https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER).

The solution used in the JS track is to use a special library to represent bigger integers. Those are often called BigInt.

Elm also has a BigInt library https://package.elm-lang.org/packages/cmditch/elm-bigint/latest/. So the simplest solution would be to import it. In that case, I feel like something should be mentionned in the readme, or a commented link to the doc in the template.

One inconvenience is that currently, dependencies also have an impact on the way build tests are run. Those are partially fixed by https://github.com/exercism/elm/pull/275 but it is not a perfect solution.

What do you think?

mpizenberg:
Hi, I'm doing some cleaning in the track before the v3 switch. This issue doesn't have a clear answer, and is getting out of date with new changes in the track so I'm going to close it.

PR #260, opened 2019-08-24
Adjust core track to help ease bottlenecks

kytrinyx:
**TL;DR:**
This PR reorders the core exercises; see the end for the new ordering.
Please read the full thing before replying!

---

I wrote a script that takes some data points into account to move certain core exercises to be optional side exercises, and which reorders the remaining exercises.

This is not meant to be the "be all end all" of perfection :-)
This is an attempt to make some simple changes in the short term. In the longer term, over the next 12 months, the product team will be doing a bunch of work to dig into what makes great Exercism exercises, and how to structure tracks to provide a better experience for both learners and mentors.

**Once this goes live** we will continue to monitor the continuation rate, the median wait time, and a few other stats as well, to help decide whether or not we need to take any other immediate steps. That said, if you see anything weird or worrying once the change is out, please ping us on Slack (or here) so we can discuss a fix posthaste!

**The most helpful thing you can do in reviewing this** is to look at the results and tell me whether anything is obviously terrible or suspiciously weird. I know very little about the Elm track (or the language, for that matter), so I could easily have missed something.

I'd also appreciate it if you posted this to the mentors as a heads up, and also because mentors tend to have a great gut sense about the core exercises.

### Removal from core:

The following exercises were moved out of the core track:

- sum-of-multiples
- accumulate

This is a list of exercises that typically are unappealing to students and not very interesting to mentor. They tend to be math problems or implementations of CS algorithms and the like.

**Worth considering:** Are any of these exercises particularly valuable in the Elm track? If so, we should keep it. Let me know and I can regenerate the PR.

### Reordering:

The data that I based the reordering on was taken from the production database, and is based on the last 6 months.

I reordered by taking the original order, and then giving "penalty points" for two different things

- low continuation rates (percentage)
- long wait times in the mentor queue (median wait time in minutes)

"Continuation" is when someone completes an exercise, but then does not submit the following exercise. We have other interesting numbers that we may use to adjust things in the future, but this one seemed like our best bet for now for the bottlenecks. I somewhat arbitrarily chose 75% as the cutoff for whether or not to give penalty points, i.e. continuation rate < 75% gets penalized. The only exercise I did not penalize based on this rate was the exercise directly following `hello-world`, since `hello-world` is basically just a taste test, and it is natural to assume that many people will decide that Exercism is not really for them.

<details>
 <summary>Continuation rates</summary>

```
- hello-world (100%)
- two-fer (95%)
- bob (57%)
- leap (43%)
- space-age (90%)
- sum-of-multiples (82%)
- accumulate (90%)
- grade-school (88%)
- raindrops (100%)
- robot-simulator (35%)
- allergies (100%)
```

</details>

Long wait times in the mentor queue is an indication that mentors don't enjoy mentoring the exercise, or put it off. This could be for a number of reasons, which we have not explicitly identified, but we've observed that sometimes the exercise is simply too difficult in the current position, which means that mentors have to go back and forth a whole bunch with students to get something right, and the discussions can be quite frustrating. Also sometimes the exercise is just not very interesting to mentor. Or if the exercise is too early in the track sometimes people will submit lots of really complicated solutions rather than a small handful of mostly reasonable solutions, also making it harder to mentor.

So this pushes exercises with long wait times further back in the core track. We may decide that we need to remove some exercises from core altogether, if wait times continue to be bad.

<details>
 <summary>Wait times</summary>

```
- hello-world (0 min)
- two-fer (1640 min)
- bob (14573 min)
- leap (1326 min)
- space-age (1886 min)
- sum-of-multiples (3000 min)
- accumulate (3728 min)
- grade-school (0 min)
- raindrops (15165 min)
- robot-simulator (0 min)
- allergies (0 min)
```
</details>

#### Before
```
- hello-world
- two-fer
- bob
- leap
- space-age
- sum-of-multiples
- accumulate
- grade-school
- raindrops
- robot-simulator
- allergies
```

#### After

_Note: the numbers indicate the position in the core track, as defined by the index of the array. Minus means that it moved earlier, plus means that it moved later._

```
- hello-world
- two-fer
- space-age (-2)
- grade-school (-2)
- allergies (-4)
- leap (+2)
- robot-simulator (-1)
- raindrops (+1)
- bob (+6)
```

I have to say that I was surprised that raindrops and leap moved to later in the track, because in other tracks these tend to be great earlier... but leap has a poor continuation rate, and raindrops has a poor wait time. If you have any thoughts about those exercise I'd love to hear them!

### DIFF
--- a/config.json
+++ b/config.json
@@ -37,6 +37,72 @@
       ]
     },
     {
+      "slug": "leap",
+      "uuid": "e943e3bf-920e-4541-95c3-be7bf6024afe",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "conditionals",
+        "integers"
+      ]
+    },
+    {
+      "slug": "space-age",
+      "uuid": "2ce5bcba-c708-4da0-ae11-4654b3e0d0a5",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "floating_point_numbers"
+      ]
+    },
+    {
+      "slug": "grade-school",
+      "uuid": "c7749c5a-529e-4e46-8bdb-e33519f6d176",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": [
+        "maps",
+        "sorting"
+      ]
+    },
+    {
+      "slug": "raindrops",
+      "uuid": "e3e73171-c3d1-49f7-9511-d1f4a08e7a69",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 2,
+      "topics": [
+        "filtering",
+        "text_formatting"
+      ]
+    },
+    {
+      "slug": "robot-simulator",
+      "uuid": "6b8f952e-94ea-4264-accf-cdb28b3b4529",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 3,
+      "topics": [
+        "records",
+        "tuples"
+      ]
+    },
+    {
+      "slug": "allergies",
+      "uuid": "29b5a28a-417a-4cee-ba6f-9dd942ceffaa",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 4,
+      "topics": [
+        "bitwise_operations",
+        "enumerations",
+        "filtering"
+      ]
+    },
+    {
       "slug": "difference-of-squares",
       "uuid": "646c7164-cfa4-42c3-9af5-2f672e6ec81e",
       "core": false,
@@ -58,17 +124,6 @@
       ]
     },
     {
-      "slug": "leap",
-      "uuid": "e943e3bf-920e-4541-95c3-be7bf6024afe",
-      "core": true,
-      "unlocked_by": null,
-      "difficulty": 1,
-      "topics": [
-        "conditionals",
-        "integers"
-      ]
-    },
-    {
       "slug": "rna-transcription",
       "uuid": "3af96d19-87d7-4d8b-aecb-f63122815501",
       "core": false,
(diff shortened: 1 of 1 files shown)


mpizenberg:
Hi @kytrinyx,

**Regarding removals**:

I agree sum-of-multiples doesn't bring much. The main insteresting concept in there is filtering lists and it is poorly emphasized in that exercise. Accumulate though shows maybe the most fundamental function for manipulating collections in elm which is the `map` function. Actually the `Accumulate` name is quite misleading because "accumulate" has another meaning usually in functional programming, and in this case this function is literally the `map` function, i.e. `accumulate = List.map` solves the exercise.

**Regarding reordering**:

space-age: Great to learn pattern matching on enumerations (called custom types in elm).

grade-school: I'm actually surprised this one arrives this early. It brings a lot of concepts at the same time, dictionaries, sorting, interop between lists and dicts. By the way, I see a 0min wait time on this exercise. What does it mean?

allergies: Very non idiomatic exercise in my opinion. It gears towards bitwise operations, that we try to avoid in elm. I see that the continuation rate of allergies is 100% and wait time is 0 min, do you also have completion rate for this exercise? I feel like this should not be in elm core exercises anyway.

leap: ok

robot-simulator: ok

raindrops: Just a bit more complex than leap year and on the same subject (divisibility) so in my opinion it should be just after leap.

bob: I agree it is a bit more complex but frankly, not too much once you get used to string manipulation functions. I feel like this one could be an early exercise if there was another simpler string manipulation exercise between two-fer and this one.

By the way, I'm not very familiar with all exercism exercises but I feel like the difficulty and core-qualified exercises are not very homogeneous. Maybe this is because I feel there hasn't been a global attempt at a coherent "curriculum" in this track. This will be a challenge for v3 of exercism.

PR #261, opened 2019-08-27
Renamed bracket-push to matching-brackets

edgerunner:
As requested in exercism/problem-specifications#1501
and detailed in exercism/elm#253


mpizenberg:
Looks good to me!

PR #273, opened 2019-10-05
Add ArmstrongNumbers exercise

CharlonTank:
Here is the elm version of Armstrong numbers exercise.


mpizenberg:
Hi @CharlonTank, I've rebased the PR on current master and moved and changed a bit the files to update it to elm 0.19.1. There was a missing test case from the problem specification for the number 0 so I've added it. Otherwise everything looks good. If I'm nitpicking, the `check` name of the function is not very informative so I've renamed it `isArmstrongNumber`.

Thank you for your contribution! Looks good to me now so I'm merging!

PS: it's usually better practice to use another branch than master in your fork for the PR. Something like the name of the exercise for example.

PR #274, opened 2019-10-05
Update npm dependencies

mkoppmann:
The current npm dependencies weren’t updated for over one year, so one currently gets this message when installing Elm:

```
found 58 vulnerabilities (8 low, 17 moderate, 33 high)
```

This PR updates `elm` and `elm-test` to the current latest version and reduces the amount of vulnerabilities to:

```
found 4 high severity vulnerabilities
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,47 +5,52 @@
   "requires": true,
   "dependencies": {
     "ajv": {
-      "version": "5.5.2",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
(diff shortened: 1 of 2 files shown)


mkoppmann:
Apparently the build fails because of the old NodeJS version. Travis is using NodeJS v6, which is end of life. Should I update NodeJS to the current LTS release v10?

mpizenberg:
Hi @mkoppmann, thank you very much for pointing this out!

I'm currently in the process of cleaning and updating the track to elm 0.19.1. I hope this won't take more than two weekends. It will effectively solve this problem so I'm closing this PR for now.

PR #275, opened 2019-10-07
Much faster bin/build.sh tests

mpizenberg:
Copy each `Tests.elm` file with the exercise suffix `Tests$exercise_name.elm`. This enables running `elm-test` once instead of for every exercise and saves a looot of time when running `bin/build.sh`.


CharlonTank:
That's amazing ;)

PR #287, opened 2019-12-04
Improve "phone-number" test coverage

SiriusStarr:
The tests covering invalid area/exchange codes for 11-digit numbers that are part of the [standard problem specification](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) were missing, leading to a hole in the test coverage..


mpizenberg:
Indeed! those tests were added later apparently (https://github.com/exercism/problem-specifications/commit/e8a51196f6fb5c0b761a1da08f1b90ade3d9bd7e). I wonder if there is a way to get notified when problem specifications change @iHiD ?

Issue #289, opened 2019-12-06
Addition of more exercises


SiriusStarr:
Is there any particular objections to adding more (side) exercises to the Elm track?  (I'm happy to do the PRs.)  I would particularly like to see "wordy" added, as there are currently no exercises aimed at `elm/parser`, which is a valuable skill for students to practice, but there are other exercises I feel could be valuable as well.

mpizenberg:
Oh you mentioning `elm/parser` reminded me of some work I had been holding off two months ago before becoming maintainer. I wanted to remove most unnecessary dependencies especially regex. You can see all the `Remove regex ...` commits there: https://github.com/mpizenberg/elm/commits/dependencies.
In particular I remember that I couldn't figure out how to do `phone-number` with just standard string functions. So I went a bit overboard in doing my own custom made parser, just for this XD https://github.com/mpizenberg/elm/commit/e800df749e5885dfda8a8bf07e630128996ee182. This definitely is a bad idea even though I don't like the regex one. Maybe using `elm/parser` for that one could make sense? Anyway, back to your question.

> Is there any particular objections to adding more (side) exercises to the Elm track?

Exercises are welcome! But I may add that if you have some time these few weeks before Christmas, Cedd and I would be super happy to get help on the research experiment matter: https://gist.github.com/mpizenberg/b686ed5723328d43266ccaff4d6fac09

Let me know by email at matthieu.pizenberg@gmail.com. Otherwise, exercises are also welcomed :)

SiriusStarr:
> Maybe using `elm/parser` for that one could make sense?

I think `phone-number` is a bad candidate for `elm/parser` since it doesn't require well-formedness of non-digits (e.g. `"  ( 5 03-----    86753    09"` should be considered a valid phone number), so using `Parser` is massive overkill for something can be solved so concisely with either pattern matching or basic string functions, haha.  I would certainly commend a student for going through the work if they went about it that way but suggest they refactor to a simpler solution.

> Exercises are welcome!

Awesome, I'll start slowly porting things over as I get bored, haha.  I will leave this issue open for a bit longer in case other people have comments.

@mpizenberg  I feel I'm lacking context for this research experiment you linked?

mpizenberg:
> something can be solved so concisely with either pattern matching or basic string functions

Well I think I wasn't able to solve it with string functions. Maybe I didn't try enough.

>  I feel I'm lacking context for this research experiment you linked?

[...]

PR #290, opened 2019-12-07
Add "wordy" exercise

SiriusStarr:
Per #289 .


mpizenberg:
The `Hints` section I think would be a good idea but I think we have to not change the generated readme, and I'm not aware of a mechanism to have a personalized section for a common exercise. Would you mind asking about this on the exercism slack?

SiriusStarr:
@mpizenberg  You can!  I just was half asleep and forgot to stage the `hints.md` file, haha.  [Here's the info on how to do it.](https://github.com/exercism/website-copy/blob/master/pages/become_a_maintainer.md#adding-hints)

mpizenberg:
> @mpizenberg You can! I just was half asleep and forgot to stage the hints.md file, haha. Here's the info on how to do it.

Oh nice, I didn't know about this one!

Regarding the implementation, I'm wondering why the `Number` type is needed. Wouldn't it be simpler to hold operations like these?

```elm
type Operation = Add Int | Multiply Int | Divide Int
```

We could have

```elm
parser : Parser (Int, List Operation)
solveProblem : (Int, List Operation) -> Int
```

I feel like it would simplify the code but not sure, I haven't done it. What do you think?

SiriusStarr:
Undoubtedly, I just copy-pasted my (admittedly crappy) Haskell solution, haha.  I will clean it up in a bit.

SiriusStarr:
Okay, updated with an even simpler solution.  Now off to fix my Haskell solution, haha.  :eyes:

mpizenberg:
Looks good!

PR #294, opened 2019-12-15
Update Word Count exercise according to specification

### LINKED ISSUE #293: Update tests of Word Count exercise
The README of the Word Count exercise states that a word will be (e.g.) a contraction of two simple words joined by a single apostrophe (ie "it's" or "they're")

However, there is no test that takes care of this special property of words.

If the tests for the Word Count exercise can be generated from data given in https://github.com/exercism/problem-specifications/blob/master/exercises/word-count/canonical-data.json, then maybe they should be regenerated. Otherwise, I can provide a PR with updated tests.

Furthermore, the example solution in `exercises/word-count/src/WordCount.example.elm` needs to be adjusted since it fails for some of the updated tests. If needed, I can include the solution I just submitted in a PR.

michaelkpfeifer:
Closes #293 

Getting rid of regular expressions in the example solution may make this a little bit more wordy.

### DIFF
--- a/exercises/word-count/src/WordCount.example.elm
+++ b/exercises/word-count/src/WordCount.example.elm
@@ -1,24 +1,42 @@
 module WordCount exposing (wordCount)
 
 import Dict exposing (Dict)
-import Regex
-import String
+
+
+sanitizeSentence : String -> String
+sanitizeSentence sentence =
+    String.toLower sentence
+        |> String.map
+            (\char ->
+                if Char.isLower char || Char.isDigit char || char == '\'' then
+                    char
+
+                else
+                    ' '
+            )
+
+
+sanitizeWord : String -> String
+sanitizeWord word =
+    if String.startsWith "'" word && String.endsWith "'" word then
+        word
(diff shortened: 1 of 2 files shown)


mpizenberg:
Looks good to me, thank you!

mpizenberg:
Actually I just had a second look at the exercise readme and it says that parsing `"That's the password: 'PASSWORD 123'!", cried the Special Agent.\nSo I fled` gives two occurences of `password`, which this implementation does not. A small fix is possible with:
```elm
sanitizeWord : String -> String
sanitizeWord word =
    if String.startsWith "'" word then
        sanitizeWord (String.dropLeft 1 word)

    else if String.endsWith "'" word then
        sanitizeWord (String.dropRight 1 word)

    else
        word
```
Do you mind if I apply this directly on master?

michaelkpfeifer:
Of course I don't mind if you apply this to master. I thought about this for a second, but I could not come up with a use case. 

Thanks for having a second look.

PR #299, opened 2020-05-06
rna-transcription: Add test for invalid input

nhawkes:
Currently it is possible to pass all tests without producing an error result which makes it confusing why the return type is result.

This changes the error type to string and enforces that *some* error is produced for invalid input.

Example failing test:
image

The example code produces the error suggested here https://github.com/exercism/elm/issues/174#issuecomment-361012095 `Result.Err "'I' is not a valid nucleotide"`

Since the task instructions do not mention any particular error message it only seems fair to accept any.

This is my first pull request here. Sorry for any issues.

### DIFF
--- a/exercises/rna-transcription/src/RNATranscription.elm
+++ b/exercises/rna-transcription/src/RNATranscription.elm
@@ -1,6 +1,6 @@
 module RNATranscription exposing (toRNA)
 
 
-toRNA : String -> Result Char String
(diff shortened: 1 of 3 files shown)


mpizenberg:
Hi @nhawkes thanks for the help! I'll try to review this during the weekend

mpizenberg:
This looks very sensible to me! Thank you @nhawkes

PR #300, opened 2020-05-11
Dependencies security fix

mpizenberg:


mpizenberg:
Travis is failing because of fetch-configlet. We can change that in another PR later. Here is the relevant info about that: https://github.com/exercism/configlet/pull/174 
I've created another issue for that specific problem in #301

ceddlyburge:
Thanks for sorting this out Matthieu, I had it on my to do list but never got round to it. I'm a bit nervious about merging code with a failing build. exercism/configlet#174 states that the problem is a transient one. I have tried rebuilding the commit, but got the same problem. Maybe we could try running the build again at a quiet time of day when the rate limit is less likely to be hit?

mpizenberg:
I've updated the `fetch-configlet` script to the current version in the configlet repo. The API endpoint and the response was different. All good now!

PR #312, opened 2021-01-07
Update template/elm.json

mpizenberg:
This fixes some issues with old versions of elm-explorations/test@1.0.0


mpizenberg:
Once this is merged, we'll have to update the test runner to pick it up: https://github.com/exercism/elm-test-runner/blob/master/Dockerfile#L29

ceddlyburge (approved):

PR #331, opened 2021-02-14
Add concept exercise for booleans

mpizenberg:
Teaches the concept of booleans, with an exercise based on [Annalyn's infiltration story][annalyn] + one function added for `if-then-else` expressions.

[annalyn]: https://github.com/exercism/fsharp/blob/main/exercises/concept/annalyns-infiltration/.docs/instructions.md

I'll still need to update the global `config.json` and the requirements for the concept exercise when we have merged the fundamentals exercise.

### DIFF
--- a/concepts/booleans/.meta/config.json
+++ b/concepts/booleans/.meta/config.json
@@ -1,0 +1,9 @@
+{
+  "blurb": "Learn how to use booleans in an Elm program",
+  "authors": [
+    "mpizenberg"
+  ],
+  "contributors": [
+    "ceddlyburge"
+  ]
+}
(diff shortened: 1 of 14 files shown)


ceddlyburge:
Hi Matthieu, this looks good to me, based on my (still coalescing) knowledge of how the concepts / exercises should be put together. Its good that the example wouldn't be obviously better represented as a Type, instead of a boolean.

mpizenberg:
I've updated this to the latest specs and now references correctly basics-2 in requirements. This should be good to merge I think.

ceddlyburge (approved):

PR #346, opened 2021-04-12
Remove 'skip' on concept exercise tests

mpizenberg:
While all except the first test of practice exercise must be skipped in the tests file, the situation is different for concept exercises. It appears we didn't notice that change few months back so I'm making the change now. The rationale is that beginner students might not be skilled enough yet to "unskip" manually the tests. So tests in concept exercises are not skipped.

This change only affects CLI users since the online test runner "unskip" everything anyway.


mpizenberg:
This fixes issue #344

ceddlyburge (approved):

Issue #347, opened 2021-04-12
Track things left to do for v3


mpizenberg:
**WIP**

## High priorities (before call for contribution)

- [x] make a todo for what's left to do for v3
- [x] remove `config/exercise_readme.go.tmpl`
- [x] remove the `reference/` directory
- [x] remove `exercises/shared/.docs/cli.md`
- [x] write the [track-specific shared files](https://github.com/exercism/docs/blob/main/building/tracks/shared-files.md#track-specific-files) `debug.md`, `help.md` and `tests.md`
- [x] add a ["foregone"](https://github.com/exercism/docs/blob/main/building/tracks/config-json.md#foregone-exercises) exercise category in config.json.
- [x] make the status of exercises "beta" and not "wip" if we want them visible for people to test
- [x] rewrite a "contributing-concept-exercise.md" document
- [x] rewrite the readme
- [ ] change concept exercises status from beta to active

## Priorities

- [ ] write as many concept + concept exercise as possible
- [ ] in `config.json` [practice exercises](https://github.com/exercism/docs/blob/main/building/tracks/config-json.md#practice-exercises), remove "topics", and fill "prerequisites" and "practices" (#343, #320)
- [x] update the `elm.json` of all exercises to the one in `template/`.
- [ ] add icon field in exercises that do not have one automatically based on "slug.svg", cf list here: https://github.com/exercism/website-icons/tree/main/v3-exercises
- [x] pick icons for key features, CF https://github.com/exercism/docs/blob/main/building/tracks/config-json.md#key-features and https://github.com/exercism/docs/issues/189

Tests runner:
  - [x] figure out how to output the code of the test
  - [ ] Add CI to prevent usage or at least warn about usage of `Fuzz`

Representer:
  - [x] integrate the work on Cedd repo into the elm-representer exercism repo

## Niceties

- [x] replace package.json by elm-tooling.json
- [ ] build analyzers
- [ ] action to help with maintenance: https://github.com/marketplace/actions/pr-commenter
- [ ] add CI to check that `elm.json` of all exercises is in sync with `template/elm.json`

Tests runner:
  - [ ] add colors

Representer:
  - [ ] evaluate the potential of normalizations based on a batch of solutions for some exercises

mpizenberg:
V3 has launched. This grouped issue is now replaced by small issues for each thing left undone here, with priorities.

PR #351, opened 2021-04-16
Remove dead wood

ceddlyburge:
As defined in #347


mpizenberg (approved):
Less things to wonder about!

PR #357, opened 2021-04-25
Remove type annotation on hello world

mpizenberg:
The hello world practice exercise is now the first thing that have to be done to unlock a track. So I just removed the type annotation on the string constant. It's not needed for hello world and introduced in basics-2.


ceddlyburge (approved):
Makes sense to keep it simple

PR #358, opened 2021-04-25
Fix long running tests

mpizenberg:
99% of the time was spent in the Pythagorean triplets test "triplets for large number" which was taking 40s. It was implemented with List methods such as `List.range`, `List.concatMap` etc. I changed the implementation to be tail call recursive and to use comparisons of the shape `x - y == 0` instead of `x == y` because of how elm optimizes those. As a result, this test now takes 1s and is still the longest running test of our suite. Meaning the practice tests suite now takes approximately 1s to run.

This was annoying me and would have been annoying for all future contributions for v3.


mpizenberg:
By the way, I'm wondering more generally, is this single test responsible for a majority of the docker running costs for exercism tests runners across tracks. Maybe @ErikSchierboom or @iHiD has some stats about that.

ceddlyburge (approved):
I'm not sure I quite understand this. It seems to me like something is wrong if the tests take 40 seconds, and maybe we could chnange the tests. Also optimised code is usually hard to read code, which is at least partially the case here, and so it would be strange to show this to a student (although maybe this is not what happens with this code).

Anyway, I expect there is probably a reason for this change that I don't understand, so I'm happy to approve it on that basis.

mpizenberg:
That test "triplets for large number" is coming from the exercise [canonical data in the problem spec](https://github.com/exercism/problem-specifications/blob/main/exercises/pythagorean-triplet/canonical-data.json#L77). The issue is that fundamentally the complexity of this exercise for a bruteforce solution is O(n^2). So when n = 30000 like in that test, the runtime explodes.

Even worse, the previous Elm solution had a O(n^2) complexity in both operations and memory since it was building list of values and filtering those lists with valid triplets. The change made the new Elm solution constant in memory since its not building intermediate lists anymore.

If you look at the [Python solution of this exercise](https://github.com/exercism/python/blob/main/exercises/practice/pythagorean-triplet/.meta/example.py) they solve it in a slightly smarter way, that is O(n * sqrt(n)) instead of n^2, which makes it roughly 170 times faster for n = 30000. The downside is that it makes the exercise more Mathy, making me think more of the type of exercises in [Project Euler](https://projecteuler.net/) than exercism.

In sum, I can see 4 ways of solving our problem:
1. Remove the exercise
2. Remove the "triplets for large number" test
3. Remove nothing and keep a bruteforce solution that is better designed for performances (tail called, constant memory)
4. Remove nothing and change the reference solution to a smarter one

I chose (3) because it was the easiest for me to do rather quickly to fix the problem but if preferable we can go with one of the other options.

ceddlyburge:
Hi Mathieu, please do go ahead and merge this if you think its the best thing to do.

I assume it would also help if we used a smaller value for N in the test? And it doesn't sound like there are any downsides to this, so probably worth doing if its in our control (which it looks like it is). So maybe we could just delete the N = 30000 test but keep the N = 840 one? 

And is the solution shown to the student at some point? If so we could show the code for various solutions (maybe the O(N^2)m the O(n log(n)) and the optimised O(N^2)) and have a few comments to discuss the trade offs.

Cheers, Cedd

mpizenberg:
Unfortunately, we don't have a `design.md` document to know what the purpose of the last test was when Katrina added it. Maybe it was specifically here to encourage students to find the smarter way to solve it instead of going the simple bruteforce way consisting of filtering all possible triplets. And maybe at that time, it wasn't a CI problem since it was only run by students.

I'll ask around in slack to see what other tracks think of that exercise.

ceddlyburge:
Makes sense. I guess in the case, the test would ideally fail if the code took too long to execute. It might be a bit fiddly to get the timing right given that there are different cpu speeds and suchlike, but the 47 seconds to 1 second difference gives quite a lot of room.

mpizenberg:
I got a wonderful example solution from the [Julia track](https://github.com/exercism/julia/blob/c6b067b39fb5217c745fba23906248a19521c7be/exercises/practice/pythagorean-triplet/.meta/example.jl) pointed to me by @SaschaMann . The tests now run in a few milliseconds and the total tests suites takes 180ms :)

ErikSchierboom:
Excellent work! Regarding the question whether this test case causes a lot of test time, I think there might be other tests that are even worse, like the large alphametics solution. That being said, the CI runs are not _that_ often that this is a big issue IMHO. Great that you've been able to optimize it for Elm!

PR #361, opened 2021-04-30
Concept maybe

mpizenberg:


ceddlyburge (approved):
Hi Matthieu, this looks good, and I would be happy to merge it.

One note is that I generally haven't been including function type signatures in the intial .elm file, and this one does have it. I'm not sure if we need to be consistent, but just flagging that we aren't.

Also, I take it that we are now planning an advanced Maybe concept? Where we can teach map and suchlike?

I wonder if it is worth teaching `withDefault` in this exercise. It is useful in the very basic 'two-fer' practice exercise, and is something that a lot of beginners don't know

Cheers, Cedd

mpizenberg:
Ha good points, I haven't had the time to check all your PR's yet, but I think except from the basics where we want students to write themselves the type signature it's good to provide a type signature for those template initial elm files. It usually makes the task clearer and also reduces potential types issues for the functions we import in the tests files. But of course we should only provides types of the exposed functions, not helper functions.

Regarding advanced concept exercises, I do thing we need few of them on most concepts like lists that I just reviewed and maybe. But you're probably right that `withDefault` could appear on that exercise without waiting on and advanced one. I think I just tried to not introduce functions from the module to focus on the core data structure. If you have an idea for introducing it here don't hesitate to update the exercise.

mpizenberg:
> I wonder if it is worth teaching withDefault in this exercise

I've added a `name : Maybe String` field to the `Player` type and added a task to write

```elm
introduce : Player -> String
introduce { name } =
    Maybe.withDefault "Mighty Magician" name
```

Don't hesitate to merge if you like it or let me know otherwise.

PR #370, opened 2021-05-25
Fix the naming of files to make configlet happy

mpizenberg:


ceddlyburge (approved):
5 passing checks, seems like configlet is happy :)

PR #377, opened 2021-06-23
Cleanup repository-specific labels

mpizenberg:
Now that exercism has defined [org-wide labels here](https://github.com/exercism/org-wide-files/blob/main/global-files/.github/labels.yml), we can use those directly so I've cleaned up most of the tags that we still had in this repo. We are only left with

- `dependencies`: used by dependabot
- `good first issue`: has special meaning for GitHub so might be useful for things like hacktoberfest
- `wontfix`: self explanatory

### DIFF
--- a/.appends/.github/labels.yml
+++ b/.appends/.github/labels.yml
@@ -3,53 +3,13 @@
 # https://github.com/exercism/org-wide-files/blob/main/global-files/.github/labels.yml.     #
 # ----------------------------------------------------------------------------------------- #
 
-- name: "bug"
-  description: ""
-  color: "fc2929"
-
 - name: "dependencies"
   description: "Pull requests that update a dependency file"
   color: "0366d6"
 
-- name: "duplicate"
-  description: ""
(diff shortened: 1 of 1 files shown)


mpizenberg:
Do you think this is reasonable @ErikSchierboom ? (I may have a tendency to want things cleaner that they need ^^)

ErikSchierboom:
@mpizenberg I'm perfectly fine with that. Each track is completely free to choose the labels they want to use. As a small side note: the Exercism-wide labels' main goal is for them to be used to present data on the website, but you are completely free to use them in your track.

Note that merging this won't yet do anything, as we don't have the labels syncing enabled yet. I'm hoping to get to that this week.

mpizenberg:
Noted thanks, I'll wait for the syncing to be operational then

mpizenberg:
Is syncing operation now?

ErikSchierboom:
@mpizenberg It is!

ErikSchierboom (approved):

PR #378, opened 2021-06-23
Use elm-tooling to remove the JS ecosystem from this repo

mpizenberg:
This will also prevent us from getting security PRs from dependabot each time some JS dependency has an issue.


ceddlyburge (approved):
This all looks good and I can see the new pipeline running and the tests passing on it.

mpizenberg:
I just switch node 10 for node 16 in the tests matrix because 10 is now EOL.

PR #381, opened 2021-07-06
Enable concept exercises

ErikSchierboom:
This PR enables the Concept Exercises for this track. Without setting this to `true`, they won't be available on the website.

Note that the Concept Exercises themselves also have to change their status from `beta` to `active` (or even better: just omit the `status` key) to enable them on the production website.

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,7 +13,7 @@
     "average_run_time": 2.0
   },
   "status": {
-    "concept_exercises": false,
+    "concept_exercises": true,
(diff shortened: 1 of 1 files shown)


iHiD (approved):

ErikSchierboom:
CC @exercism/elm for changing the Concept Exercise status from `beta` to `active`

mpizenberg:
Thanks, will do it as soon as I come back from holiday :)

PR #382, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/src/HelloWorld.elm
+++ b/exercises/practice/hello-world/src/HelloWorld.elm
@@ -1,6 +1,5 @@
 module HelloWorld exposing (helloWorld)
 
 
-helloWorld : String
 helloWorld =
     "Goodbye, Mars!"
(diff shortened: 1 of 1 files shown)


mpizenberg (commented):
That's cleaner indeed! and the "Goodby, Mars!" makes it clear what they have to change. We probably don't need the type annotation though since they are not mandatory and introduced in the second concept exercise. What made you feel we should add the annotation?

ErikSchierboom:
@mpizenberg Oh no real reason, I just used the example file as the basis, and that apparently includes the annotation. We should definitely lose the annotation if it is not needed. I'll update!

exercises/practice/hello-world/src/HelloWorld.elm
```diff
@@ -1,22 +1,6 @@
...
-
-   The module name is expected by the test program and must match the name of this
-   file. It has to stay just the way it is.
--}
-
-
 module HelloWorld exposing (helloWorld)
 
 
+helloWorld : String
```
ErikSchierboom, line 4:
```suggestion
```

mpizenberg (approved):
Awesome, thanks @ErikSchierboom !

PR #391, opened 2021-07-25
concept: generics

mpizenberg:


exercises/concept/story-name/.meta/design.md
```diff
@@ -0,0 +1,28 @@
...
+
+- Understand what a type variable is.
+- Know how to define generic types.
+- Know how to define generic functions.
+- Introduce higher order functions.
+
+## Out of scope
+
+- Covariance and contravariance.
+
```
ceddlyburge, line 17: Maybe add Never, Phantom Types and Opaque types as out of scope, although it would probably be useful to mention them in passing so that students know they exist

exercises/concept/story-name/.meta/design.md
```diff
@@ -0,0 +1,28 @@
+# Design
+
+## Goal
+
+The goal of this exercise is to teach students how to consume and make generic code.
+
+## Learning objectives
+
+- Understand what a type variable is.
```
ceddlyburge, line 9: It might be worth calling out `Cmd Msg` and `Cmd msg` somewhere, which is something that confuses a lot of people new to elm I think.

exercises/concept/story-name/.docs/introduction.md
```diff
@@ -0,0 +1,27 @@
...
+
+- as input
+- as output
+- as both input and output
+- parametric polymorphism VS ad hoc polymorphism VS subtyping polymorphism
+
+## Higher order functions
+
+- basics of higher order functions: dependency injection.
+- common patterns accross container types (map, andThen, etc.)
```
ceddlyburge, line 27: Maybe add this to the learning objectives?

exercises/concept/story-name/.docs/introduction.md
```diff
@@ -0,0 +1,27 @@
+# Introduction
+
+Generics are a way of writing code that avoid duplications of code.
```
ceddlyburge, line 3:
```suggestion
Generics are a language feature that enable code reuse and advanced type patterns (which are explained in a later concept).
```

exercises/concept/story-name/.meta/design.md
mpizenberg: Right! that's an important remark!

exercises/concept/story-name/.meta/design.md
mpizenberg: Well for Never, I'm not sure we even need to mention it since it's not really a generic, just a standard type that you have no way to construct. Phantom types, I'm not sure they are out of scope, at least introducing them. That will probably depend on if the concept grows a bit big already without them. As for opaque types, they are not linked to generics in any specific way, they are just normal types. The only thing special is that you don't export the types internal so it's more of public/private and module-related concept.

ceddlyburge: Yep, that is true, although in my head they are all in the "advanced types" bucket :)

mpizenberg:
I've filled it and updated this with your remarks. I've also removed the section about higher order functions since it actually does not have anything to do with generics. This is a higher order function: `foo : (Int -> Int) -> String` for example.

Let me know what you think of the updated content!

exercises/concept/story-name/.docs/introduction.md
```diff
@@ -1,27 +1,141 @@
...
+In the definition of `Distance` above, `unit` is a free type parameter, not bound to any data in the type.
+We can also call this technique **phantom types**.
+It is surprisingly useful when we want to enforce constraints at compile type.
+For example, we want to make sure that we only add distances of the same unit.
+
+```elm
+-- The module does not expose the internals of Distance,
+-- but instead two constructor functions: meter and foot.
+-- We do not want to expose 'Distance(..)' because we do not want
+-- people outside this module to access the internals of Distance.
```
ceddlyburge, line 104: I might word this bit slightly differently. The main thing with opaque types is that you can't create an instance of a type directly, you have to use one of the exported creator functions.

ceddlyburge (approved):
Looks good :)

PR #393, opened 2021-07-26
Concept: pattern matching

mpizenberg:


ceddlyburge (commented):
Generally I think that pattern matching is quite specific to the context (lists, tuples, etc), and might be better introduced within that context. Maybe a 'wrapping up' type of thing might be useful as well, I"m not sure.

mpizenberg:
> Generally I think that pattern matching is quite specific to the context (lists, tuples, etc), and might be better introduced within that context.

Makes sense. I'll write a complete `introduction.md` and we'll see what can be ported to the relevant context concept exercises, if all can be or if it makes still sense to have some of it left in a separate concept as this.

mpizenberg:
Alright, I've completed the introduction document with all the things I had in mind regarding pattern matching. I personally see a lot of value in having all this in a single document, but let me know if you still think otherwise. By the way, since repetition is a good way to learn, I don't think it's bad if we have some repetition between specific pattern matching examples in dedicated data structure concepts and in this document.

Maybe we could call the concept "Advanced pattern matching and destructuring" or something like that, and just let the first section be a reminder of things already seen in the custom type concept?

ceddlyburge (approved):
Hi Matthieu, this looks good. I have some fairly general comments

Pattern matching only happens inside a `case` statement I think. It would probably be worth mentioning this, and stating that pattern matching *is* a case statement

It looks like it might be difficult to create an exercise for this concept.

I think it is probably worth mentioning where this concept overlaps with previous ones (custom types, tuples, records, maybe others)

There is no example of tuple pattern matching I think

It is probably worth pointing out code that is non idiomatic (but still useful for demonstration purposes)

Maybe the example code should adhere to elm-format

There is probably scope for more list pattern matching examples (I quite like this: https://www.elm.christmas/2020/8)

jiegillet:
Hello!

I actually came up with a nice exercise idea for this concept. I have written the code, but I still need to write the tests and instructions.

Is it cool if I take over?
If so, what's the best way? A PR to the `pattern-matching` branch?

mpizenberg:
I'm fine with anyone taking over. By the way, some of the feedback there on reddit interesting to read, just as a data point for exercises https://www.reddit.com/r/elm/comments/rvdz3p/comment/hrp4a19/?utm_source=share&utm_medium=web2x&context=3

PRing to the branch is a possibility. I also wouldn't mind to add you to the members able to commit in exercism/elm but that would need to be cleared also by Cedd and Erik (and I can't actually do it, so would need Erik anyway).

mpizenberg:
Another option is we close this PR and you open a new one with the same contents and some context linking to this.

jiegillet:
Thank you! I'll keep working on it then, and I'll take a look at the link!

> PRing to the branch is a possibility. I also wouldn't mind to add you to the members able to commit in exercism/elm but that would need to be cleared also by Cedd and Erik (and I can't actually do it, so would need Erik anyway).

I'd actually like to become a maintainer at some point, but I don't feel like I deserve it yet, haha :)

I'll probably branch this PR and open a new one then.

PR #395, opened 2021-08-01
Tuples concept and associated exercise

ceddlyburge:


mpizenberg:
I'm starting the review here. Regarding the last commit about icons, would you mind moving it to PR #394 ? We can merge that one first to get the other PRs passing the checks then.

ceddlyburge:
I think there is no need to move the commit, it turns out that it is making the same changes.

exercises/concept/tisbury-treasure-hunt/.docs/introduction.md
```diff
@@ -0,0 +1,81 @@
+# Tuples
+
+Tuples in Elm can hold two or three values, and each value can have any type, although idiomatically the two value tuple is most commonly used.
```
mpizenberg, line 3:
```suggestion
Tuples in Elm can hold two or three values, and each value can have any type.
```

mpizenberg (approved):
Thanks @ceddlyburge , this looks great to me! I did some small fixes and changes with a commit for each. And I have a suggestion to change the first sentence cause I don't agree with 2-tuples being more idiomatic than 3-tuples. To me it's mostly a matter of needs, and I would use `(a, (b, c))` where I can use `(a, b, c)`.

The only thing left is adding to `config.json` I think.

ceddlyburge:
> I don't agree with 2-tuples being more idiomatic than 3-tuples

Hi Matthieu, this is mostly based on the fact that Tuple.first and second exist, but not third. And mapFirst and mapSecond exist, but not mapThird.

Also [the tuple package](https://package.elm-lang.org/packages/elm/core/latest/Tuple) has an explicit example of using a record instead of a 3 tuple, in the very small amount of description that it has.

So I think it is worth mentioning that two tuples are favoured over 3 tuples. Would you prefer some different words to do this?

mpizenberg:
In my opinion the Tuple module in elm/core is misnamed and should be named
Pair. It's not as much as Tuple favors 2-tuple but rather that Evan only
provides helper functions for Pairs (which the module should be renamed to)
because they happen so often. I'd rather say that only pairs have built-in
helper functions in elm/core rather than saying they are more idiomatic.
What do you think of this wording?

By the way, regarding the example saying that 3-tuples should use records
instead, I'm convinced Evan added that in the API to justify the fact that
tuples longer than 3 are not possible anymore. The general justification is
that it's always better to have named fields than to have unnamed ones.
That's also true for pairs. But removing tuples all together would have
been a bit too much.

ceddlyburge:
Hi Matthieu, it would be super trivial to add Tuple.third and mapThird. Either of us could create a PR for that in under 10 minutes I think! So I think that it must be a deliberate design choice, to encourage the use of 2 tuples over 3 tuples, and as part of the effort to generally prefer records over tuples.

I generally prefer records to tuples, but they do still come in handy (I used them extensively in the elm-representer, and they are a core part of the state monad for example).

So how about making a statement that explains the Elm way of thinking, in place of the current statement.

So we could remove this: `although idiomatically the two value tuple is most commonly used`

And add something like this: `In Elm, Tuples are mostly used for simple transient types. For this reason Tuples have a maximum of 3 values, and as a further encouragement convenience functions are only supplied for 2 values. For more complex data, it is best to switch to records. If your record contains a bunch of Bool and Maybe values, you may want to use to Custom Types`

mpizenberg:
Actually would not be trivial since it would require another module.
`Tuple.first` can only work on 2-tuples because its type is `(a,b) -> a`
and that does not work on 3-tuples. So if he wanted to add built-in helper
for 3-tuples, it would either require new names, like `Tuple.firstOfPair`
and `Tuple.firstOfTrio`, or would require another module entirely like
`Trio.first`, `Trio.second`, `Trio.third` etc. So that's code duplication
and additional names to find. And since Evan alway chooses minimalism, it's
better to only have helpers for 2-tuples which are indeed much more common
than 3-tuples.

> In Elm, Tuples are mostly used for simple transient types. For this
reason Tuples have a maximum of 3 values, and as a further encouragement
convenience functions are only supplied for 2 values. For more complex
data, it is best to switch to records.

That sounds quite nice. Maybe (pun intended) leave the last sentence about
Bool and Maybe out?

ceddlyburge:
Yep, you would need Triple.first etc.

Seems like we are both happy with this text though, so lets go with that

> In Elm, Tuples are mostly used for simple transient types. For this reason Tuples have a maximum of 3 values, and as a further encouragement convenience functions are only supplied for 2 values. For more complex data, it is best to switch to records.

PR #405, opened 2021-09-03
Remove text in concept exercise requirements

mpizenberg:
We have been writing the "Concept" and "Requirements" parts of the `design.md` doc differently for the concept exercises. This change consists in simplifying and standardizing that section of this document.

Basically it now simply contains the list of the hashes of concepts required. There is no need to repeat what is already said just above or to detail requirements that are pretty obvious from the concept name. Elixir is doing the same.


ceddlyburge (approved):
Yep, this is better, and it will avoid inconsistencies. The documentation format still seems strange to me, with a lot of similar / repeated information.

PR #420, opened 2021-09-17
That might solve the URL not showing up as one

leojpod:
When viewing the instructions on the exercism website, the link appears as plain text.
I assume that the Markdown renderer does not check for "implicit" links, unlike the GitHub preview.


mpizenberg (approved):
Thanks! I remember seeing other places with links not showing. We'll probably have to do a full check of all markdown documents with links to catch them all.

leojpod:
or patch the markdown engine 😆

leojpod:
Damn, the platform is quick! 🥇 
it already added the points!

mpizenberg:
Yep, Jeremy and Erik did a fine job ;)

PR #441, opened 2021-11-12
Corrected pattern matching call to list head

pwadsworth:


mpizenberg (approved):
Thanks for the fix @pwadsworth !

PR #444, opened 2021-11-26
Add bowling exercise from leojpod

ceddlyburge:


ceddlyburge:
Hi @leojpod, I've updated this exercise to the v3 format and added another example solution. I'm afraid I couldn't easily push to your branch so I've created a new pull request. You could just check this out and then create a pull request from your repo if you want, in which case I will close this one.
And you are still down as the "author" of the exercise in the config.json
Can you give it a quick review and then I'll ask Matthieu to take a look and then we can merge it in.
Cheers, Cedd

leojpod:
Hi @ceddlyburge, 

Looks good :) Thanks for that! 
I'll create it right away

PR #445, opened 2021-11-29
Bowling exercise updated by ceddlyburge

leojpod:


leojpod:
@ceddlyburge There it is :)

mpizenberg:
Thanks @ceddlyburge and @leojpod ! I can review this wednesday I think

mpizenberg (commented):
I love bowling so this fun to review ^^. I have a few remarks on things that would be good to change in my opinion.

In the tests, when using `Expect.equal` with a pipe, it's better to use the forward pipe since it corresponds to the error messages reported by the tests runner.
For example, if we have the following test:

```elm
Expect.equal (Ok 0) <|
    score [ 1, 0, ..., 0 ]
```

The error message will be:
```txt
↓ Bowling Score
✗ should be able to score a game with all zeros

    Ok 1
    ╷
    │ Expect.equal
    ╵
    Ok 0
```
with 1 and 0 reversed.

In the instructions, I see that there is a mention of 2 functions `roll` and `score` however only `score` is to implement in practice. And the code syntax is not Elm so we should revise a bit this `Requirements` section of the instructions.

The `score : List Int -> Result String Int` function is returning a result, but the error side of the result is never used in the tests, just the fact that it errored. In my opinion, we should either use the error in the tests or get rid of it and return a `Maybe Int` instead. In case we want to use the error cases, expecting a string would be a bad idea for UX because people will just spend time fixing their string messages to match the one in the tests. So we might instead predefine a custom type in the template for the errors. But that would need to be well done such that no test could potentially return to different valid errors because otherwise we would fail the tests if a student does not choose the same one as the tests. All things considered, it's better to start with `Maybe` and later improve the exercise if we think it's useful.

I've added a third implementation since it was a bit of fun. For the default example, I'd suggest picking one of example2 or example3 instead of the current one because of it's "pyramid of doom"-like approach, but I'd let you decide what to do.

ceddlyburge:
HI @leojpod, I'm happy to make these changes if you prefer (I feel bad that we ignore your PR for so long). If that works for you, can you give me push access to your repo?
Thanks, Cedd

leojpod:
Hi @ceddlyburge, 
That would be great thanks, I'm not in my most efficient phase right now and I'd really appreciate your help :)

mpizenberg:
Hi Cedd, as you're one of the maintainer of this repo, you can already push
to Leo's branch as I did. You just have to add Leo's fork as a remote. And
push with the identifiers you use for the exercism repo

ceddlyburge:
Hi @mpizenberg , I've made these changes, can you take another look?
Thanks, Cedd

mpizenberg (approved):
Thanks @leojpod and @ceddlyburge !
Looks good to me.

leojpod:
Thanks a lot to the two of you!

PR #451, opened 2022-02-04
Sync: update `bowling` with configlet

jiegillet:
I ran `configlet sync` on the track and saw a bunch of things that could be fixed, so I'd like to work through these if it's OK.
Some syncs involve adding new tests, I'll make separate PRs for those, starting with this one.

For the `bowling` exercise:
* synced `instructions.md` (automated)
* added `tests.tom` (automated)
* added the `skip <|` in the tests to match the template
* added one new test ("last two strikes followed by only last bonus with non strike points") from latest specs in problem-specifications. Sorry it's impossible to notice in the diff because of the added `skip <|`.
* synced `config.json` (automated + I added my name as a contributor since I added a test)

### DIFF
--- a/exercises/practice/bowling/.meta/config.json
+++ b/exercises/practice/bowling/.meta/config.json
@@ -1,10 +1,11 @@
 {
-  "blurb": "Score a bowling game",
+  "blurb": "Score a bowling game.",
   "authors": [
     "leojpod"
   ],
   "contributors": [
-    "ceddlyburge"
+    "ceddlyburge",
+    "jiegillet"
   ],
   "files": {
     "solution": [
@@ -19,6 +20,6 @@
       ".meta/src/Bowling.example.elm"
     ]
   },
-  "source": "Inspired by Uncle Bob's Bowling Game Kata",
+  "source": "The Bowling Game Kata at but UncleBob",
   "source_url": "http://butunclebob.com/ArticleS.UncleBob.TheBowlingGameKata"
 }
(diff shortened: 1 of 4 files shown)


exercises/practice/bowling/.meta/config.json
```diff
@@ -19,6 +20,6 @@
       ".meta/src/Bowling.example.elm"
     ]
   },
-  "source": "Inspired by Uncle Bob's Bowling Game Kata",
+  "source": "The Bowling Game Kata at but UncleBob",
```
ceddlyburge, line 23: Hi Jie, this text looks wrong, was it done by configlet?

If so I think we should update configlet, if not then I think the original text is ok.

Cheers, Cedd

jiegillet: It was configlet, yes.
You're right, it doesn't look correct, but that's the data over on [problem-specs](https://github.com/exercism/problem-specifications/blob/main/exercises/bowling/metadata.toml), it would need to be changed over there.

jiegillet: Actually, it DOES look correct:
http://butunclebob.com/ArticleS.UncleBob.TheBowlingGameKata

"but UncleBob" is the name of the site.

ceddlyburge: Well, it doesn't look right! Maybe at butunclebob.com would be better. I'll merge this and create an issue on problem-specs ....

ceddlyburge (approved):

PR #461, opened 2022-02-12
Add concept `pattern-matching` and concept exercise `ticket-please`

jiegillet:
Supersedes #393.

Realted to @ceddlyburge's comment [in here](https://github.com/exercism/elm/pull/393#pullrequestreview-723717695) about `about.md`, I tried formatting the example code with `elm-format` but it made it way too long, students will get discouraged if they need to scroll so much. 

Please have a look, and I'll be happy to discuss.


exercises/concept/ticket-please/.docs/hints.md
```diff
@@ -0,0 +1,47 @@
...
+- You can destructure a custom type with a single variant `type Box = Box AB` directly in a function argument like so: `(Box ab)`.
+- You can pattern match the field `a` of a record `{a : A, b : B}` using: `{ a }`.
+- You can use both the above techniques at the same time for custom types that hold a record.
+- You cannot directly pattern match the pair `ab` from a record `{ ab : (A, B) }`. You must pattern match `{ ab }` first, then use `( a, b) = ab` in a `let` binding.
+- To find the number of elements that satisfy a condition in a list, use [`List` functions][list-docs].
+- You can destructure pairs diectly in anonymous functions arguments.
+
+## 3. Detect tickets assigned to the Dev Team
+
+- You can destructure a custom type with a single variant `type Box = Box AB` directly in a function argument like so: `(Box ab)`.
```
ceddlyburge, line 26: A lot of these hints are repeated in the various sections, maybe they could go in the General section?

exercises/concept/ticket-please/.docs/instructions.md
```diff
@@ -0,0 +1,90 @@
...
+Use `List` functions to do the count and use destructuring in the function argument of an anonymous function to seperate the user/comment pair.
+
+```elm
+Ticket
+    { status = Closed
+    , createdBy = ( User "Chiaki", 1 )
+    , assignedTo = Nothing
+    , comments = [ ( User "Chiaki", "Never mind, fixed it." ) ]
+    }
+    |> numberOfCreatorComments
```
ceddlyburge, line 40: Is the  `|>` operator taught as a prerequisite of this exercise?

exercises/concept/ticket-please/.meta/Exemplar.elm
```diff
@@ -0,0 +1,67 @@
+module TicketPlease exposing (..)
+
+
+type Ticket
```
ceddlyburge, line 4: I think it would be better to have these already defined for the user, probably in a supporting file. The Go exercise has an example of this (https://github.com/exercism/elm/tree/main/exercises/concept/go/src)

exercises/concept/ticket-please/.meta/Exemplar.elm
```diff
@@ -0,0 +1,67 @@
...
+type User
+    = User String
+
+
+emptyComment : ( User, String ) -> Bool
+emptyComment ( _, comment ) =
+    String.isEmpty comment
+
+
+numberOfCreatorComments : Ticket -> Int
```
ceddlyburge, line 30: I'm not sure this task adds anything that isn't already done in other tasks? So could probably be removed.

exercises/concept/ticket-please/.meta/Exemplar.elm
```diff
@@ -0,0 +1,67 @@
...
+
+        Just (User "Charlie") ->
+            True
+
+        _ ->
+            False
+
+
+assignTicketTo : User -> Ticket -> Ticket
+assignTicketTo user (Ticket ({ status } as ticket)) =
```
ceddlyburge, line 58: Nice demonstration of `as`

exercises/concept/ticket-please/tests/Tests.elm
```diff
@@ -0,0 +1,277 @@
+module Tests exposing (task1, task2, task3, task4)
+
+import Expect
+import Test exposing (Test, describe, test)
+import TicketPlease exposing (..)
+
+
+task1 : Test
```
ceddlyburge, line 8: I think there is a way that the test runner wants these tests to be defined, and I think it is to have them all in the same function, and use `describe` to separate out the tests for the various cases. Also, just for information, the test runner doesn't support use of fuzz testing (which isn't an issue for the PR, but useful to know)

exercises/concept/ticket-please/tests/Tests.elm
```diff
@@ -0,0 +1,277 @@
...
+            \() ->
+                Ticket { status = New, createdBy = ( User "Jeremy", 5 ), assignedTo = Just (User "Charlie"), comments = [] }
+                    |> assignedToDevTeam
+                    |> Expect.true "Expected ticket assigned to Charlie from dev team"
+        ]
+
+
+task4 : Test
+task4 =
+    describe "Task 4: TicketsPlease.assignTicketTo"
```
ceddlyburge, line 106: Some of these tests are quite long, and could be made shorter and more expresser with some helper functions I think.

Something along the lines of this, which removes the irrelevant information and highlights the information that is important to the test.

```
test "assign new, unassigned ticket to Roy" <|
            \() ->
                { vanillaTicket : createdBy = user "John" }
                    |> assignTicketTo (User "Roy")
                    |> Expect.equal
                       { vanillaTicket : createdBy = user "John" }
                            , assignedTo = user "Roy"
                        )
```

ceddlyburge (commented):
Thanks Jie, this looks good, and great to get another concept / exercise ready. These really are the big thing about v3, and we still have a way to go on the Elm track (although that is also true of most of the other tracks as well).

Im also a fan of these blog posts, which might be good in the links
https://www.bekk.christmas/post/2020/8/peeking-inside-lists
https://www.bekk.christmas/post/2020/7/simplify-your-code-with-ad-hoc-tuples

mpizenberg:
I haven't looked at all, but I have one remark on the code organization. So I believe we tried to standardize tests files by adding a description corresponding to the test file, so there needs to be a `TicketPlease` (careful no `TicketsPlease`) at the top like:
```
tests : Test
tests =
    describe "TicketPlease"
        [ task1, task2, ...]
```
Then I remember we have implemented in the tests runner a way to group tests by tasks according to their description. I think the description had to be exactly the same as the task in the exercise description, Or it had to be a number like "1" or something. The details are fuzzy for me now since we did it quite some time ago. @ceddlyburge will know I think.

Also to be able to generate the automatic extraction of the tests code as required by the test runner spec, Cedd did some elm parsing work that got integrated in the test runner. Meaning the tests code must follow some specific guideline to be parsed correctly. Again I don't remember the exact details and Cedd will know. It's possible that you cannot split the tests in different top level functions but have to put all in a single nested test with describes.

Other than this remark, I'll let Cedd and others comment on the theme and implementation of the exercise.
Thanks for all you work on this!

exercises/concept/ticket-please/tests/Tests.elm
jiegillet: That's a good idea. I had helper functions at the beginning but then remembered that concept exercises tests are not shown, so tests should be as self contained as possible. Your approach however, is much better than mine :)

jiegillet (commented):
Thank you both for the super quick reviews, I didn't expect that.

@ceddlyburge I'll work those links in! 
@mpizenberg , I'll look up the specks for the tests, thanks for letting me know.

jiegillet:
I looked at the test-runner code and comments on PR, as far as I could gather, a specific structure (`describe "ExerciseName"`) is required, and extra nesting of `described` is possible, but I don't think there is any code or syntax to identify tasks. I've updated accordingly.

mpizenberg:
I found the part of the test runner that extract the task id: https://github.com/mpizenberg/elm-test-runner/pull/9/files
Basically it needs to be a description level above the tests grouped together, which only contains the number of the task.

It appears it was done for version 2.0 of elm-test-rs, it seems I didn't brought it yet in the exercism test runner: https://github.com/exercism/elm-test-runner/blob/main/Dockerfile#L23

So that's another thing that will need updated. But at least you know that in future we will be able to have tests grouped by task if a description with the single task number is grouping them. I think I didn't prioritize this integration because it wasn't rolled out yet in exercism. I haven't followed lately so I don't know what happened on exercism side and if now tests can be grouped by tasks in the UI

jiegillet:
OK I see, so once this gets merged, we will need a `describe "1"` to group task 1 tests. I think it's not quite worth doing it here, all tests will need to be adapted anyway. I'll be happy to help with that when the time comes.

> I haven't followed lately so I don't know what happened on exercism side and if now tests can be grouped by tasks in the UI

Coincidentally, I saw yesterday a [thread on Slack](https://exercism-team.slack.com/archives/CARRG4MNC/p1644600581040059) about the weekly called that mentioned "We're showing off the new task_id stuff in the editor right now ". I didn't attend the call (it's like 2AM for me) but I'm guessing it's imminent.

exercises/concept/ticket-please/.meta/Exemplar.elm
ceddlyburge: Yep fair enough. This was a comment Erik left me so I'm passing it on. I think the concern is that a user might edit the code, and could then get in a muddle, and having it in a separate file makes it very clear that the code shouldn't be edited.

ceddlyburge:
The tests look like they will work with the test runner now. The [tests for the test runner](https://github.com/exercism/elm-test-runner/tree/main/extract-test-code/tests) are probably the best place to look for whats supported at the moment, but we should add some better documentation for that.

Do you want to let us know when you have finished making changes and we can take another look?

Thanks! Cedd

jiegillet:
Yes, I did refer to the tests in the test-runner. I agree, documentation would be nice.

I also didn't find a way to run the test-runner easily without docker (which I'm not familiar with yet). Maybe we could add a script that does that? `run_locally.sh` or something. Or possibly, if we could include a CI check here that tests the exercise on the test-runner, that would be cool (I don't know how to do that though haha).

With my last commit, I've done all I wanted to do, we can investigate.

ceddlyburge:
Yep, these are all good ideas. Running the test runner with docker is actually pretty easy (and is a more thorough test, as this is what happens on the website), but you do need to have docker installed. Maybe we can give it a try on our video call later.

exercises/concept/ticket-please/.docs/introduction.md
```diff
@@ -0,0 +1 @@
+Copy from concepts/pattern-matching/introduction.md after approval.
```
ceddlyburge, line 1: Hi Jie, can you copy this across now?

exercises/concept/ticket-please/.meta/Exemplar.elm
ceddlyburge: Hi Jie, sorry to be a pain, but I think we should do this. It's a suggestion from Erik, and we have already done it for other exercises, so I think it is best to be consistent. We could maybe bring it up at a weekly call sometime, and see if we can address your point about the user experience being worse. Then we could apply the change to all the exercises at once to keep the consistent experience.

ceddlyburge (approved):

exercises/concept/ticket-please/.docs/introduction.md
jiegillet: Done!

jiegillet:
I've split the solution and added a support file, but I ran into some complications. `build.sh` was assuming there there would be only one file, or take the first one alphabetically, so that broke things for my files `TicketsPlease.elm` and `Support.elm`. 

I modified `build.sh` to take the `exercise-name` from the path and make it into `ExerciseName` instead. But then I had to rename the files in the `Go` exercise, since in there `Go.elm` was the support file. 

Then I realized that the file names must be unique because `build.sh` brings them all together and compiles it in one go. So I came up with the names `TicketsPlease.elm` and `TicketsPleaseSupport.elm` because it doesn't sound too bad for a convention.

Is that cool?

PR #475, opened 2022-03-26
Rewrite `robot-simulator` tests that the Elm test runner can't parse

jiegillet:
Solves [this issue](https://github.com/exercism/elm-test-runner/issues/33). 
Copy of the explanation: 

> [The test runner] code extractor fails on the current tests in `describe "right"` and `describe "left"`, because of this line
> 
> ```elm
> |> List.indexedMap (\i e -> test ("step " ++ String.fromInt i) (\() -> e))
> ```
> 
> The code extractor cannot work with `test ("step " ++ String.fromInt i)` it can only deal with a single string literal (like `test "step 1"`). Changing the code extractor to deal with this would be very complex, and even if it did manage to extract the code, it would display `e`, which is fairly useless, so I think we should rewrite those tests instead.

This should solve the problem. I opted to remove `scanl` for simplicity, but I can put it back if it's easier to read. My weapon of choice would have been `List.Extra.iterate` for this one.

### DIFF
--- a/exercises/practice/robot-simulator/tests/Tests.elm
+++ b/exercises/practice/robot-simulator/tests/Tests.elm
@@ -1,4 +1,4 @@
-module Tests exposing (assertionList, tests)
+module Tests exposing (tests)
 
 import Expect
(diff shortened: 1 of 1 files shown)


mpizenberg (approved):
Much better!

PR #478, opened 2022-03-27
Apply `elm-review` suggestions to Example/Exemplar solutions

jiegillet:
I ran `NoUnused` and `Simplify` rules on all the example and exemplar solutions and found a bunch of suggestions.
This fixes all of them (except one that was wrong, I've already reported it and it will get removed).

### DIFF
--- a/exercises/concept/tracks-on-tracks-on-tracks/.meta/Exemplar.elm
+++ b/exercises/concept/tracks-on-tracks-on-tracks/.meta/Exemplar.elm
@@ -1,6 +1,6 @@
 module TracksOnTracksOnTracks exposing (..)
 
-import List as List
+import List
 
 
 newList : List a
--- a/exercises/practice/acronym/.meta/src/Acronym.example.elm
+++ b/exercises/practice/acronym/.meta/src/Acronym.example.elm
@@ -9,7 +9,7 @@ abbreviate phrase =
     phrase
         |> Regex.split dashOrSpace
         |> List.map firstLetter
-        |> String.join ""
+        |> String.concat
 
 
 firstLetter : String -> String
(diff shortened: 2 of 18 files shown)


mpizenberg (approved):
The changes look good to me thanks!

Is it possible/difficult to also add those checks in the CI scripts? Or would it be more maintenance burden if things start breaking (new rules reporting unwanted errors)?

jiegillet:
I can think of a couple of ways to have these kinds of checks in the CI.

1. Add `elm-review` directly to the repo and CI, exactly like in `elm-analyzer`. This would only work for common rules, not the specialized rules for the exercises, but that's all we need for checking new exercises anyway. This would need a bit of manual maintenance (adding/removing rules match the analyzer rules), but not too bad. It would be easy to control, because we can suppress checks for specific files if need be.

2. Find a way to run the production `elm-analyzer` directly from the CI. I've asked Erik about that for `elm-test-runner` because it would be neat, but he hasn't responded yet. I don't know if it's feasible. This would require no maintenance on the rules, but we would lose a bit of control.

3. Clone `elm-analyzer` in the CI and run cheks from there (or use a git submodule if we want to be able to run the checks locally).

If I were to make a new exercise, I think I would like to have it reviewed locally first, so either 1 or 3 (with submodule option) would be neat, all you would need is running `elm-review --config ../../..` (or absolute path to `elm`/`elm-analyzer` repo) from the exercise folder.

mpizenberg:
I think I like option (1) consisting in adding elm-review to this repo. Checking common rules is enough in my opinion. And if we want a single source of truth for common rules, I believe we could even publish a package like `elm-review-exercism-common` and use it both for the analyzer and for the main track repo.

In any case, those changes are good so we can merge them now and open a new PR for setting up elm-review for CI or local+CI.

PR #480, opened 2022-04-02
Create a practice exercise generator

jiegillet:
I know that this is not a priority compared to concept exercises, but I love implementing practice exercises, and I had some time off, so I decided to build a generator to help creating those.

The generator code is in `generate`, all you need to make it work is running `bin/generate.sh <exercise-name>` and follow the instructions.

For example, try to run `.bin/generate.sh dominoes`, it will ask you to modify `config.json` and after that the following files will be created:
```
exercises/practice/dominoes/
├── .docs
│   └── instructions.md
├── .meta
│   ├── config.json
│   └── src
│       └── Dominoes.example.elm
├── elm.json
├── src
│   └── Dominoes.elm
└── tests
    └── Tests.elm
```

the docs and config file are created by configlet, and the rest by the generator. Once the solution is implemented for real, the instructions will ask you to run one more configlet command to generate the `tests.toml`.

The Elm files generated are by no means perfect, but it really gets the ball going because all of the data is there. With some good regex-fu, the tests are not too hard to get into shape. For examples for `dominoes` I feel it would be enough to replace the lists `[1, 2]` by tuples `(1, 2)` and start working on the solution, it can be done with one clever `sed` command.
I tried to catch some corner cases (all sorts of input, reserved words, reimplemented exercises) but it's probably not going to be perfect for all exercises. I've ran it on a bunch and it seems to work well enough though.

The code is not really critical to anything, so I feel that a deep review may not be necessary here, but of course suggestions are welcome.

I decided to test it on a real case: `complex-numbers`, but if that's too much to review, I can put it in a different PR.

As with all large PRs, I do not expect a quick review, especially since there are other ones pending.


mpizenberg (commented):
I must say I wasn't expecting an elm cli app to generate templates. My main objection is that elm is not suited for this task. This is a lot of template filling, network requesting, json parsing. My feeling is that doing all this directly in python or javascript will be much more straightforward and clearer, making it easier to maintain.

I don't want to discard all you work, and I totally get wanting to do it in elm, though I'm feeling that there is nothing fundamentally requiring elm for this (we don't need to parse elm files as I understand) and elm is not very suited for this kind of tasks. So I'm wondering if the most adequate way for this isn't instead a markdown readme containing a few commands reusing the files in `template/` and filling the data with appropriate data from the canonical repo.

What do you think?

Also I think it's worth splitting this PR and the one for the creation of the complex exercise.

jiegillet:
It's true that this is well suited for a scripting language... 

The logic in `generate/src/Main.elm` is not completely straightforward though, for example `searchFunctions`, and I think that more than a few commands would be necessary would be required to transform `canonical-data.json` into a functional `Test.elm`.

But if you think this is a maintenance burden, I understand and I can close the PR. I'll definitely keep using it for myself on my branch :)

mpizenberg:
I'll have a closer look at this in the coming days, and probably give a try at a JS equivalent, to see the tradeoffs.

mpizenberg:
I've started some review of the code but it would be nice if you could add a bit more comments on how things work. I've added some comments and some remarks directly in comments and I have not spend enough time yet to understand it all.

Some questions I had while reading, why is `reimplements` per test and not per exercise? and what is the purpose of `replaceReservedWord`? I didn't get why there could be elm keywords in the canonical value decoded but I might see it from the wrong perspective.

mpizenberg:
Ok so for `reimplements` it's because tests should be immutable so when changing a test case we can track which test a new one is re-implementing.

mpizenberg:
I've added `set -euo pipefail` because I got in a confusing situation where I had answered `yes` to he question without actually changing the `config.json` (I thought I didn't need if I didn't change anything). Then the `configlet sync` command failed, and so the directory for the folder was not created, and so the `cp template/elm.json $exercise_dir` create a file of the name of the exercise instead of inside it. Basically a cascade of errors.

With the fail fast, I saw the error with `configlet sync` saying the exercise was missing from the config file and changed that.

mpizenberg:
Also there might be some issue in the escaping of strings for function parameters in the generated test file since when run with `accumulate`, I'm seeing this in the generated test file:

```elm
Accumulate.accumulate [ "a", "b", "c" ] "(x) => accumulate([" 1 ", " 2 ", " 3 "], (y) => x + y))"
    |> Expect.equal [ [ "a1", "a2", "a3" ], [ "b1", "b2", "b3" ], [ "c1", "c2", "c3" ] ]
```

Where the `"` inside the accumulate expression are not escaped and thus split the string into multiple strings.

generate/src/Main.elm
```diff
@@ -0,0 +1,423 @@
...
+generate the code for that test case.
+-}
+printTest : String -> Dict String ( Function, List String ) -> Case -> String
+printTest exercise functions testCase =
+    let
+        addReimplementsExplanation : Maybe String -> Maybe String
+        addReimplementsExplanation =
+            Maybe.map
+                (\uuid ->
+                    "This exercises reimplements another exercise with uuid "
```
mpizenberg, line 213: Here all `exercise` should be replaced by `test` in the comment description.

generate/src/cli.js
```diff
@@ -0,0 +1,26 @@
...
+
+// required to write files
+const fs = require('fs')
+
+// Required to run HTTP requests (usually handled by browser)
+global.XMLHttpRequest = require('xhr2');
+
+// Link to compiled Elm code main.js and run it
+var Elm = require("./main").Elm;
+var main = Elm.Main.init({ flags: slug });
```
mpizenberg, line 12: Instead of only passing the slug to the Elm worker, I'd suggest downloading the canonical data first in JS and passing both the slug and canonical data in the worker. This avoids both the need for the xhr2 dependency (+ package.json and package lock) and the webdata elm dependency to remove a bunch of code. You can pass the downloaded json as a `Json.Decode.Value` in the flags and decode it with the same decoder already in your elm code.

mpizenberg:
I've also started / made an attempt at a JS equivalent of the elm script. Don't hesitate to have a look and to let me know what you think. I'll try to finish it tomorrow.

generate/src/cli.js
jiegillet: Yes, I agree, it would be cleaner.

jiegillet:
> what is the purpose of `replaceReservedWord`? I didn't get why there could be elm keywords in the canonical value decoded but I might see it from the wrong perspective.

It's for when variable names from the inputs are Elm reserved words. It actually happened to me during a test, I think the variable was called `type` or something and the file wouldn't compile/format. Those are printed in the stub file. 

> Also there might be some issue in the escaping of strings for function parameters in the generated test file since when run  [...] Where the " inside the accumulate expression are not escaped and thus split the string into multiple strings.

Good catch!

jiegillet:
> I've also started / made an attempt at a JS equivalent of the elm script. Don't hesitate to have a look and to let me know what you think. I'll try to finish it tomorrow.

I don't actually know JS (big gap in my knowledge, I'll fill it some time) so I won't be much help :)

mpizenberg:
> I don't actually know JS (big gap in my knowledge, I'll fill it some time) so I won't be much help :)

Nice actually I'm happy to hear that there are some elm coders that don't know JS :)
And by the way I'm even more interested in hearing from you if you find that JS code readable and easy enough to follow.

jiegillet:
I read through it and yes, I would say it's quite readable and I could follow.
Of course, it seems to be written in a pretty functional style, so that helps a lot :)
Of course of course, it follows my own code, so that helps to follow as well, haha :)

generate/generate.js
```diff
@@ -0,0 +1,253 @@
...
+module Tests exposing (tests)
+
+import ${exercise}
+import Expect
+import Test exposing (Test, describe, skip, test)
+
+${toElmComments(canonicalData.comments)}
+
+tests : Test
+tests = describe "${exercise}" [ ${allTestCode} ]
```
jiegillet, line 64: One thing that's missing is removing the first `skip <|` in the test file.

generate/generate.js
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  // Generate a re-implement message if this tests re-implements another one
+  let reimplements = [];
+  if (testCase.reimplements) {
+    reimplements = [
+      "This test reimplements the test with uuid " + testCase.reimplements,
+      "Please identify that test and remove it. Link:",
+      "https://github.com/exercism/problem-specifications/blob/main/exercises/" +
+        testCase.reimplements +
```
jiegillet, line 138: This value here should be the original slug, not the uuid

generate/generate.js
```diff
@@ -0,0 +1,253 @@
...
+    .join(", ");
+
+  const allFunctionsCode = Object.entries(extractedFunctions)
+    .map(([key, value]) => generateFunctionCode(key, value))
+    .join("\n\n");
+
+  const mainFile = `
+module ${exercise} exposing (${functionList})
+
+${allFunctionsCode}
```
jiegillet, line 44: One way your script is superior to mine is that it keeps the order of the functions as they appear in the canonical data (I think). I store mine in a `Dict`, so the are printed in alphabetical order.

generate/generate.js
```diff
@@ -0,0 +1,253 @@
...
+        return "Nothing";
+      } else if (json.constructor === Array) {
+        // Apply a recursive call to all elements of the array
+        const arrayContent = json.map(jsonValueToElm).join(", ");
+        return `[ ${arrayContent} ]`;
+      } else {
+        // Apply a recursive call to all attribute values,
+        // and lowercase the object keys.
+        const toElmKey = (key) =>
+          replaceReservedWord(key[0].toLowerCase() + key.slice(1));
```
jiegillet, line 184: Setting the first character to lower case is a good call! I haven't seen variables with uppercase, but it doesn't mean there aren't any.

generate/generate.js
```diff
@@ -0,0 +1,253 @@
...
+tests = describe "${exercise}" [ ${allTestCode} ]
+`;
+
+  console.log(testFile);
+}
+
+// Generate the template code for one function
+function generateFunctionCode(name, { args, canError }) {
+  const returnType = canError ? "Result String Todo" : "Todo";
+  const typeAnnotation = [...Array(args.length).fill("Todo"), returnType].join(
```
jiegillet, line 73: What does the `...` notation do? I know I should google it haha :D

generate/generate.js
mpizenberg: Yes, adding all the skips is not something that I found useful when I was adding an exercise. I wanted to discuss this with you by the way. In the template, I've added a commented skip `-- skip <|` so that people can run `elm-test-rs` directly in the exercise directory.

My workflow was to first develop the solution in the `src/Main.elm`, run the tests while writing the solution. And when the solution is done, copy it to the example and remove the implementatiin from the template.

Considering this workflow, it was easier to be able to run the tests directly. Then when I've finished, uncomment all the `-- skip` except for the first one.

Is your workflow different ? Do you run `bin/build.sh` instead of directly run the tests of just the current exercise yourself?

generate/generate.js
mpizenberg: Woops, good catch

generate/generate.js
mpizenberg: Ah I had not thought of this. Keeping the order is rather interesting because usually the order makes sense and we want to keep it.

generate/generate.js
mpizenberg: Yes I saw that when trying the script on the nucleotide count canonical data where the object keys are uppercase "A", "G", "T", "C".

generate/generate.js
mpizenberg: It's just syntax to concatenate an array with other values easily

```
let a = [ 1, 2, 3 ]
let b = [ ...a, 4, 5 ]
```

generate/generate.js
jiegillet: Oh, I see, that's rather clever. I guess I also uncomment the skips first, and add them back later. This saves a step.
Yes, I have the same workflow as you.

As an aside, I really don't get the Exercism thing of skipping tests. I really don't think it's that overwhelming to get many failing tests at first. First thing I always do for any exercise is removing the skips. Ah well...

mpizenberg: > As an aside, I really don't get the Exercism thing of skipping tests.

I'm on your side too. But I remember discussions happening on that subject and skipping was the chosen decision so we comply with the rules ;)

mpizenberg:
I've made some updates to `generate/generate.js` to match the equivalent of the elm impl. I think we have ironed out most of potential issue with that script and so I'd love to have some feedback from other people with elm knowledge about this PR, especially @ceddlyburge and @ErikSchierboom if they have the time.

Here is a summary of what this PR brings and how it's implemented.

The goal if this PR is to ease the addition of practice exercises from the problem specifications repo (with canonical data) to the elm track. This PR proposes two implementations for this functionality, that should be roughly equivalent in features. The first implementation is in bash + Elm, the second is in bash + JS. And the bash part is the same for both, so for both the command would be the same:

```sh
bin/generate.sh <exercise-slug>
```

This script will

- provide a chunk of json to be embedded in the `config.json` by the user of the script
- sync the exercise with configlet
- download the canonical data json
- call the script (Elm / JS) with the slug as argument and canonical data provided through stdin
- that script will generate the main template file as well as the tests file

For each script (Elm version / JS version) the steps are roughly the following:

- extract the list of functions to implement from canonical data. This also extracts the arguments names, and identify if the function may error if some expected result contains an error.
- generate the main template file thanks to the extracted functions
- reproduce the hierarchy of tests in canonical data into a hierarchy of tests in the `tests/Tests.elm` file
- for each tests, arguments are extracted from canonical data and converted as closely as possible to equivalent Elm values
- same for the expected values
- comments and potential reimplementations are also added as comments just before corresponding tests

Now both script implementations have advantages in drawbacks. The Elm script is composed of mainly elm code so that could be easier to maintain by maintainers not familiar with JS, however the Elm script requires an additional step to be compiled to JS in the bash script, and is composed of 3 files (elm.json 24 loc, cli.js 25 loc, Main.elm 429 loc). The JS script has the drawback of needing someone familiar enough with JS to maintain, but is rather small at 257 loc compared to Elm version.

ceddlyburge:
Ok, I've taken a look! Personally I am happy with either Javascript or Elm. There are pros and cons, but to me, it's not a huge deal either way.

For me, the pros of the Elm solution are: I enjoy writing Elm more than I like writing Javascript, the json parsing is more robust, and the types improve readability and error messages.

And I think the pros of the Javascript are: it is shorter (although not by that much, as Elm code is famous for large vertical spacing), JSON is Javascript so parsing it is easy, IO is easy, no extra step is required in the script.

So I would be happy with either, and will leave the decision to people that have a stronger opinion one way or the other ...

generate/elm.json
```diff
@@ -0,0 +1,24 @@
+{
```
ceddlyburge, line 1: Could we rename this directory to make it clear what is being generated? And maybe add a readme if we feel that it is warranted? It could have some instructions for how to generate a new practice exercise from the canonical data for example.

ceddlyburge (approved):

mpizenberg:
Thanks @ceddlyburge . I'm slightly in favor of the JS version, but since this is @jiegillet proposition and he's certainly the one going to use it most, let's leave the decision to Jeremie then since you don't have preferences Cedd.

If we go with the Elm version, I'd like to give another review pass and try some refactoring (refactoring is my favorite hobby)

jiegillet:
Well if you leave it to me, I guess I'll pick the Elm, because I love Elm, I don't know JS (yet), I'm probably going to use it a lot (and most likely modify it), and finally who am I to deprive @mpizenberg from his favorite hobby?

I'll leave it in your hands then, refactor away and I'll review it.

ceddlyburge:
Refactoring is very satisfying :)

mpizenberg:
I've made the following changes to the Elm version:

1. Replace `resultType : ResultType` by `canError : Bool` in the `Function` type. Generally it's better to use custom types over booleans, especially when these booleans are just identified by position in a list of arguments. But here, the boolean has a clear attribute name in a record, and the use of booleans enables simpler logic two combine them in the search function. As for the `canError` name. I thought it reflected better the actual data, because it might also be relevant to use `Maybe` instead of `Result` in some cases (maybe in the future).

2. Use `Dict.insert` instead of `Dict.update` in the search function. The `Dict.update` has a rather cumbersome api, and it's usually simpler to just insert, with a default value that's the current one.

3. Use multiline strings templates, with `String.replace` for `printFunction` and `printTest` as those are much easier to read.

4. Delay the path generation to the JS code. This enables doing it in an OS-independent way and help with windows compatibility if someone writes a windows equivalent to the bash script.

5. Split the decoding and the string conversion of JSON values in two steps. This enables simpler code for both the decoder and the string conversion to Elm code.

mpizenberg:
There is one last thing I want to try is another custom type for canonical data.

mpizenberg:
Ok I only did a tiny custom type change for test cases since what I had in mind didn't pan out.

With the few changes I did regarding the decoding and string conversions, the order in which all functions are in the file may be a bit untidy, feel free to rearrange as you like. I'm done with the modifications I wanted to try, the ball is in your court now.

generate_practice_exercise/src/Main.elm
```diff
@@ -45,7 +45,8 @@ type alias TestData =
 
 
 type alias Function =
-    { arguments : List String
+    { arguments : List ( String, String )
```
mpizenberg, line 48: A record will avoid possible order confusion I think `List { argName : String, argType : String }`

mpizenberg (commented):
Nice addition to have a better guess at types! I think you could even do the same for the arguments types. Using multiple tests to guess better the types. Since tests usually start with trivial tests, it's highly probable that first tests will have arguments like `[] : List todo` or `null : Maybe todo` that could be refined with the other tests using the same function, as you do for the return type.

bin/generate_practice_exercise.sh
```diff
@@ -0,0 +1,53 @@
...
+exercise_dir="exercises/practice/${SLUG}"
+
+# build configlet
+echo "Fetching latest version of configlet..."
+./bin/fetch-configlet
+
+# Preparing config.json
+echo "Please adapt and add the following in config.json before continuing: "
+UUID=$(bin/configlet uuid)
+jq -n --arg slug "$SLUG" --arg uuid "$UUID" \
```
ErikSchierboom, line 22: You could use `jq` to add the new entry to the `config.json` file. See https://stackoverflow.com/a/42248841

bin/generate_practice_exercise.sh
```diff
@@ -0,0 +1,53 @@
...
+echo "Press 'y' to continue..."
+read -n 1 k <&1
+echo ""
+if [[ $k != 'y' && $k != 'Y' ]] ; then
+    echo "Exiting"
+    exit 0
+fi
+
+# Create instructions and config files
+./bin/configlet sync --update --yes --docs --filepaths --metadata --exercise "$SLUG"
```
ErikSchierboom, line 34: Nice use of `configlet sync`!

bin/generate_practice_exercise.sh
```diff
@@ -0,0 +1,53 @@
...
+cp template/elm.json $exercise_dir
+
+# Build generator
+echo "Building exercise generator..."
+cd generate_practice_exercise
```
ErikSchierboom, line 39: Just in case you didn't know, you can use `pushd` and `popd` to temporarily change directories: https://en.wikipedia.org/wiki/Pushd_and_popd

bin/generate_practice_exercise.sh
```diff
@@ -0,0 +1,53 @@
...
+elm make --debug src/Main.elm --output=src/main.js
+cd ..
+
+# Create Elm files
+echo "Creating Elm files..."
+mkdir -p ${exercise_dir}/tests ${exercise_dir}/src ${exercise_dir}/.meta/src
+curl https://raw.githubusercontent.com/exercism/problem-specifications/main/exercises/${SLUG}/canonical-data.json \
+    | node generate_practice_exercise/src/cli.js $SLUG
+elm-format --yes ${exercise_dir}/**/*.elm
+elm-format --yes ${exercise_dir}/.meta/src/*.elm
```
ErikSchierboom, line 49: Wouldn't these filesd be matched by the previous line too?

mpizenberg: The `**` patern doesn't go in hidden directory to my knowledge so I think both are needed

ErikSchierboom: Ah, okay

generate_practice_exercise/src/Main.elm
```diff
@@ -335,16 +356,39 @@ jsonValueToElmCode json =
...
+                [] ->
+                    "List todo"
+
+                head :: _ ->
+                    "List (" ++ jsonValueToElmType head ++ ")"
+
+        JsonObject obj ->
+            case obj of
+                _ ->
+                    List.map (\( key, val ) -> cleanVariable key ++ " = " ++ jsonValueToElmType val) obj
```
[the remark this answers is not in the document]
jiegillet: Oops! Well, `elm-format` takes care of that anyway ;)

bin/generate_practice_exercise.sh
jiegillet: For some reason, when I do `**` it doesn't match hidden directories. Maybe it's a config thing on my side, but I'd rather leave it this way if others are in the same case.

bin/generate_practice_exercise.sh
jiegillet: That seems obvious, I don't understand why I haven't done that already :)

jiegillet: Ah, I remember now. It's a pain to do in-place replacement, it's not a feature of `jq`. I'll figure it out :)

jiegillet:
Thank you for all the reviews, improvements all around.

I renamed the script and folder to `generate_practice_exercise`, removed the now redundant `stub-new-exercise.sh`, added type annotation, which `JsonValue` made possible (it's not perfect by any means but it's good enough for now), improved the bash script and added a README. I'll merge it as it is, I'll probably make some changes as I use it in the future.

This has been really, really fun, thank you all. Hopefully this is a useful tool.

mpizenberg:
Thank you! I hope I haven't been too picky even if you said it was fun, I know I can be a bit too much sometimes ^^.

PS, we can safely remove the `generate.js` in a following PR

PR #481, opened 2022-04-02
Tweak debug message

jiegillet:
As discussed [here](https://github.com/exercism/elm-test-runner/pull/36), I never fully understood the message about `Debug.log` until @mpizenberg explained it in detail, so I tried rewriting it.

I've spent too much time reading it, re-reading it and rewriting it and the words don't make sense to me anymore, so it's time to ask for your help :)

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -1,21 +1,21 @@
 # Debug
 
-When a test fails, a message is displayed describing what went wrong and for which input. You can also use the fact that the tests will report any log output. You can write to the log using [Debug.log](https://package.elm-lang.org/packages/elm/core/latest/Debug#log)
-
-There is a slight nuance in that log output for constants / functions without a parameter are essentially optimised out by the transpiler, and so are not captured.
-
-In this case the function has a parameter so everything works as expected
+You can use [Debug.log](https://package.elm-lang.org/packages/elm/core/latest/Debug#log) to attach messages to failing tests, like so:
 
 ```elm
+preparationTimeInMinutes : Int -> Int
 preparationTimeInMinutes numberOfLayers = 
     Debug.log "preparationTimeInMinutes" 2 * numberOfLayers
 ```
 
-In this case the function doesn't have a parameter (and so is essentially a constant, as functions are pure), and the log output is not captured.
+On a failing test, the attached log will be `preparationTimeInMinutes: 2`.
+
+There is one caveat: top-level functions without parameters (constants) are evaluated only once before tests start running, therefore tests cannot capture their log messages. For example:
 
 ```elm
+expectedMinutesInOven : Int
 expectedMinutesInOven = 
-    Debug.log "expectedMinutesInOven" 40
-
+    Debug.log "expectedMinutesInOven" 2 * 20
 ```
 
+will not log anything during tests because the JavaScript variable associated to `expectedMinutesInOven` has already been evaluated to the value 40.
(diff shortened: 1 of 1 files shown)


exercises/shared/.docs/debug.md
```diff
@@ -1,21 +1,21 @@
...
 ```elm
+preparationTimeInMinutes : Int -> Int
 preparationTimeInMinutes numberOfLayers = 
     Debug.log "preparationTimeInMinutes" 2 * numberOfLayers
 ```
 
-In this case the function doesn't have a parameter (and so is essentially a constant, as functions are pure), and the log output is not captured.
+On a failing test, the attached log will be `preparationTimeInMinutes: 2`.
+
+There is one caveat: top-level functions without parameters (constants) are evaluated only once before tests start running, therefore tests cannot capture their log messages. For example:
```
mpizenberg, line 13: Here is an alternative wording, let me know what you think
```suggestion
There is one caveat: calls to `Debug.log` evaluated when the code is loaded instead of when the tests are run cannot be captured and reported in the tests messages. This is the case when called directly in constants definitions such as bellow
```

I'm saying it here this way, because `Debug.log` can be reported even when appearing in the code of a constant, if that code is not executed, like when defining lambdas:
```elm
constantWithLambda =
  { identity = \x -> Debug.log "x" x
  , answer = 42
  }
```
Though I'm not sure if the nuance is needed when the target of this message are people having difficulties with the language

exercises/shared/.docs/debug.md
```diff
@@ -1,21 +1,21 @@
...
+You can use [Debug.log](https://package.elm-lang.org/packages/elm/core/latest/Debug#log) to attach messages to failing tests, like so:
 
 ```elm
+preparationTimeInMinutes : Int -> Int
 preparationTimeInMinutes numberOfLayers = 
     Debug.log "preparationTimeInMinutes" 2 * numberOfLayers
 ```
 
-In this case the function doesn't have a parameter (and so is essentially a constant, as functions are pure), and the log output is not captured.
+On a failing test, the attached log will be `preparationTimeInMinutes: 2`.
```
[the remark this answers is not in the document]
jiegillet: One of my goals here was to actually log a constant to undermine my previous misunderstanding of "you can't log constants", but yes, we can change the function. How about

```elm
hoursToSeconds : Int -> Int
hoursToSeconds hours =
  let
    secondsInOneHour = Debug.log "secondsInOneHour" (60 * 60)
  in
  hours * secondsInOneHour
```

This is "complex" enough to justify logging a (constant) intermediate value.

We can change the second function to be

```elm
secondsInOneHour : Int
secondsInOneHour =
  Debug.log "secondsInOneHour" (60 * 60)
```

exercises/shared/.docs/debug.md
jiegillet: I didn't think of that, yeah, it gets pretty nuanced...

But I feel that "This is the case when called directly in constants definitions" doesn't work because it can still be understood as "constant definitions in a let block", which is the main misunderstanding I want to avoid.

How about "top-level functions without parameters that return constants are ...". It's still a simplification because lambdas can be called constants in FP, but at least it's entering a grey area and beginners wouldn't think of it that way... Maybe? :)

exercises/shared/.docs/debug.md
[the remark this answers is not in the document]
mpizenberg: I very much prefer this new hours and seconds example!

exercises/shared/.docs/debug.md
mpizenberg: We should probably go with the simplest most probable case and just ignore rare cases of like lambdas within constants.
> There is one caveat: top-level constants are evaluated only once before tests start running, therefore tests cannot capture their log messages. For example:

jiegillet:
Ok, I've updated according to the comments.

mpizenberg (approved):

mpizenberg:
Thank you @jiegillet it's much clearer now!

PR #482, opened 2022-04-04
New practice exercise: `complex-numbers`

jiegillet:
Split from #480.

### DIFF
--- a/exercises/practice/complex-numbers/tests/Tests.elm
+++ b/exercises/practice/complex-numbers/tests/Tests.elm
@@ -1,0 +1,241 @@
+module Tests exposing (tests)
+
+import ComplexNumbers as Complex exposing (Complex)
+import Expect exposing (Expectation, FloatingPointTolerance(..))
+import Test exposing (Test, describe, skip, test)
+
+
+tolerance : FloatingPointTolerance
+tolerance =
+    Absolute 0.00001
+
+
+approximatelyEqual : Complex -> Complex -> Expectation
+approximatelyEqual z1 z2 =
+    Complex.sub z1 z2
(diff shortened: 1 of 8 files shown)


exercises/practice/complex-numbers/tests/Tests.elm
```diff
@@ -0,0 +1,241 @@
...
+approximatelyEqual : Complex -> Complex -> Expectation
+approximatelyEqual z1 z2 =
+    Complex.sub z1 z2
+        |> Expect.all
+            [ Complex.real >> Expect.within tolerance 0
+            , Complex.imaginary >> Expect.within tolerance 0
+            ]
+
+
+tests : Test
```
ceddlyburge, line 22: Does this work with the test runner? I think it does but can't remember all the details, we should probably add some documentation somewhere!

And does it format the test results in a nice way on the website? I also can't remember where we are at with grouping tests in headings to make this better.

ceddlyburge (approved):
Hi Jeremie, thanks for this, it looks good. Apologies for the slow review, I didn't get an email about it, which seems strange as I do have notifications turned on. Anyway, its reviewed now :)

exercises/practice/complex-numbers/tests/Tests.elm
jiegillet: Yes, it works fine.
Grouping the tests with the headings is only for concept exercises, this one is a practice exercise, so there is nothing special in the UI. It's also OK to have helper functions like above because the test file is fully visible in the UI.

jiegillet:
Thank you for the review, don't worry about the timing, I certainly don't :)

PR #494, opened 2022-07-02
Update hints.md

Fadhil:
To get a list of all second values for a list of tuples we would do
`List.map Tuple.second list_of_tuples`


ceddlyburge (approved):

ceddlyburge:
Thanks Fadhil, good spot :)

Fadhil:
Thank you!

PR #495, opened 2022-07-08
importing dict should be here by default

gacallea:
Importing Dict should be here by default, at least. Doing so also avoids errors if running tests without solutions (i.e: will all Debug.todo still in place)

### DIFF
--- a/exercises/concept/top-scorers/src/TopScorers.elm
+++ b/exercises/concept/top-scorers/src/TopScorers.elm
@@ -1,5 +1,6 @@
 module TopScorers exposing (..)
 
+import Dict exposing (Dict)
 import TopScorersSupport exposing (PlayerName)
 
 


ceddlyburge:
@mpizenberg  @jiegillet what do you think?

I think this is the first time a concept exercise has required an import.

It does seem like it would be helpful to avoid frustration, but it is also doing a bit of the exercise for the student.

Adding a comment stating that the import is required could also work.

gacallea:
ohps, I didn't think about the exercise aspect of it! I'd say you have a point.

ceddlyburge:
It can be really frustrating learning a language though, especially when starting with a blank page and nobody to ask and no code to copy paste! So it would be good to make the learning process as smooth as possible I think. I'll let the other admins have their say and then we can make a decision. Probably we should have a policy, such as "All example code should compile and run, but fail all the tests" or something like that ...

mpizenberg:
We should definitely strive for something like all example code should compile and run, but I don't think making it a hard rule is necessary as it may be difficult depending on a specific concept goals.
As for this concept, my point of view is that we are passed the point of teaching imports so having the import already in the file is probably a good idea.  I'm not the one who designed this exercise though so I'll let you guys decide what to do.

mpizenberg:
Uh if it is indeed the first exercise that has a hard requirement for an import (previously was only checked by the analyzer right?) then maybe we add a first task consisting in importing the dict module ? no need for a function to write, it just resolves iself when the file compiles ?

ceddlyburge:
I think, on balance, it is probably best to supply the import statement as part of the template. The import concept is already taught as part of the 'basic' exercises. I think dict is currently the first exercise to require it, but it won't necesarily always be, as students can choose their path through the exercises.

ceddlyburge (approved):

PR #496, opened 2022-08-12
nucleotid-count: Convert tests and samples to new specification Fixes #483

### LINKED ISSUE #483: Nucleotide Count - Instructions doesn't fully correspond to the task
I solved the exercise by relying more on the type annotations of the `nucleotideCounts` function (`nucleotideCounts : String -> NucleotideCounts`) and on the tests than on the instructions themselves, which say "If the string contains characters that aren't A, C, G, or T then it is invalid and you should signal an error." I check also a couple of the community solutions but no one of the few I checked is also implemented to signal an error, cause also this would generate an error since the expected output is NucleotideCounts . Not sure if this is on purpose.

NyuB:
Proposal to fix [this issue](https://github.com/exercism/elm/issues/483)

The thread suggested to use Maybe to represent an error, but since the canonical test cases indicate an error message I thought it would be more accurate to use Result.

Since this is the first time I contibute to Exercism, I apologize if i missed some rules or guidance for this kind of PR, and am ready to correct my mistakes according to your remarks.

### DIFF
--- a/exercises/practice/nucleotide-count/.meta/config.json
+++ b/exercises/practice/nucleotide-count/.meta/config.json
@@ -5,6 +5,7 @@
   "contributors": [
     "albronca",
     "nathanielknight",
+    "NyuB",
     "parkerl",
     "tuxagon",
     "ulve"
--- a/exercises/practice/nucleotide-count/.meta/src/NucleotideCount.example.elm
+++ b/exercises/practice/nucleotide-count/.meta/src/NucleotideCount.example.elm
@@ -11,34 +11,37 @@ type alias NucleotideCounts =
     }
 
 
-init : NucleotideCounts
-init =
-    { a = 0
-    , t = 0
-    , c = 0
-    , g = 0
-    }
-
-
-update : Char -> NucleotideCounts -> NucleotideCounts
+update : Char -> NucleotideCounts -> Result String NucleotideCounts
 update char counts =
     case char of
         'A' ->
-            { counts | a = counts.a + 1 }
+            Ok { counts | a = counts.a + 1 }
 
         'T' ->
-            { counts | t = counts.t + 1 }
+            Ok { counts | t = counts.t + 1 }
 
         'C' ->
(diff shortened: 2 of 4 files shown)


mpizenberg:
Hi @NyuB thanks for taking you time to contribute!
If you want to run the test suite locally, you can run `bin/build.sh`

NyuB:
Thank you, I confused myself moving code between  .meta/src and src/

mpizenberg (approved):
Awesome, thanks @NyuB !
I have one remark, regarding the `propagateOrUpdate` function. Usually, when you see a function that does not modify the error side, but does something in the ok side, it means you can use `Result.map` or `Result.andThen`. Here is typically a case where `andThen` fits perfectly, and instead of writing the `propagateOrUpdate` function, you could have written:
```elm
    String.foldl (Result.andThen << update) (Ok init) sequence
```
I'm fine either way but I'm interested in hearing your point of view as well as @ceddlyburge .

NyuB:
Totally agree.

For the record, initially i was looking for a "fold with break" function like this to avoid traversing the entire String when encountering an error
```
foldbr : (a -> b -> Result c b) -> b -> List a -> Result c b
foldbr f init l =
    case l of
        [] ->
            Ok init

        h :: t ->
            let
                r =
                    f h init
            in
            Result.andThen (\v -> foldbr f v t) r
```

mpizenberg:
Ah yeah there is no fold with break in elm standard library. You need to
write your own recursive function when you want one.

ceddlyburge:
Hi both, I'm away at the moment, I'll take a look on Monday when I get
back. It sounds good though!

ceddlyburge (approved):
This looks good to me. The fold with break point is an interesting one, but I think unnecessary in this case. Result.andThen will skip the 'update' after the first error is encountered, so will just be looping through a relatively short string, so in this example problem I don't think there is a need to manually optimise (and you could imagine an implementation where all the errors are returned and not just the first one). Obviously there will be situations where it makes sense though.

PR #498, opened 2022-08-25
Rename Luigis Luscious Lasagna to Marios Marvellous lasagna

### LINKED ISSUE #497: The Lasagna exercise is duplicated
Witness: [search on "lasagna"](https://exercism.org/tracks/elm/exercises?criteria=lasagna).

Not necessarily a problem &ndash; especially as their foci are actually different &ndash; but the extremely similar names did confuse me.

ceddlyburge:
Fixes #497, and makes the two exercises more distinct when searching, so
that it is less confusing and harder to mistake one for the other.


ceddlyburge:
@mpizenberg @jiegillet can you take a look when you have a minute?
Thanks, Cedd

mpizenberg (approved):
Good idea to make the two more distinguishable. Thanks!

PR #501, opened 2022-09-14
Add practice exercise `sgf-parsing`

jiegillet:
Sorry I've been inactive for a while :)
I got the itch to add an exercise, so here goes.

This one was the absolute hardest to do in Elixir because there is no parsing library in the standard library (that's all you get in Exercism). However with `elm/parse`, it was an absolute pleasure.

### DIFF
--- a/exercises/practice/sgf-parsing/.docs/instructions.md
+++ b/exercises/practice/sgf-parsing/.docs/instructions.md
@@ -1,0 +1,66 @@
+# Instructions
+
+Parsing a Smart Game Format string.
+
+[SGF](https://en.wikipedia.org/wiki/Smart_Game_Format) is a standard format for
+storing board game files, in particular go.
+
+SGF is a fairly simple format. An SGF file usually contains a single
+tree of nodes where each node is a property list. The property list
+contains key value pairs, each key can only occur once but may have
+multiple values.
+
+An SGF file may look like this:
+
+```text
+(;FF[4]C[root]SZ[19];B[aa];W[ab])
+```
+
+This is a tree with three nodes:
+
+- The top level node has three properties: FF\[4\] (key = "FF", value
+  = "4"), C\[root\](key = "C", value = "root") and SZ\[19\] (key =
+  "SZ", value = "19"). (FF indicates the version of SGF, C is a
+  comment and SZ is the size of the board.)
+  - The top level node has a single child which has a single property:
+    B\[aa\].  (Black plays on the point encoded as "aa", which is the
+    1-1 point).
+    - The B\[aa\] node has a single child which has a single property:
+      W\[ab\].
+
+As you can imagine an SGF file contains a lot of nodes with a single
+child, which is why there's a shorthand for it.
+
+SGF can encode variations of play. Go players do a lot of backtracking
+in their reviews (let's try this, doesn't work, let's try that) and SGF
+supports variations of play sequences. For example:
+
+```text
+(;FF[4](;B[aa];W[ab])(;B[dd];W[ee]))
+```
+
+Here the root node has two variations. The first (which by convention
+indicates what's actually played) is where black plays on 1-1. Black was
+sent this file by his teacher who pointed out a more sensible play in
+the second child of the root node: `B[dd]` (4-4 point, a very standard
+opening to take the corner).
+
+A key can have multiple values associated with it. For example:
+
+```text
+(;FF[4];AB[aa][ab][ba])
+```
+
+Here `AB` (add black) is used to add three black stones to the board.
+
+There are a few more complexities to SGF (and parsing in general), which
+you can mostly ignore. You should assume that the input is encoded in
+UTF-8, the tests won't contain a charset property, so don't worry about
+that. Furthermore you may assume that all newlines are unix style (`\n`,
+no `\r` or `\r\n` will be in the tests) and that no optional whitespace
+between properties, nodes, etc will be in the tests.
+
+The exercise will have you parse an SGF string and return a tree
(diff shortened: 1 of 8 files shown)


exercises/practice/sgf-parsing/.docs/instructions.md
```diff
@@ -0,0 +1,66 @@
...
+```text
+(;FF[4];AB[aa][ab][ba])
+```
+
+Here `AB` (add black) is used to add three black stones to the board.
+
+There are a few more complexities to SGF (and parsing in general), which
+you can mostly ignore. You should assume that the input is encoded in
+UTF-8, the tests won't contain a charset property, so don't worry about
+that. Furthermore you may assume that all newlines are unix style (`\n`,
```
ceddlyburge, line 59:
```suggestion
UTF-8 and the tests won't contain a charset property. Furthermore you may assume that all newlines are unix style (`\n`,
```

exercises/practice/sgf-parsing/.meta/src/SgfParsing.example.elm
```diff
@@ -0,0 +1,144 @@
...
+
+
+sgf : Parser Tree
+sgf =
+    Parser.oneOf [ sgfWithParentheses, sgfWithoutParentheses ]
+
+
+sgfWithParentheses : Parser Tree
+sgfWithParentheses =
+    Parser.succeed identity
```
ceddlyburge, line 44: I haven't got my head fully in to this code, but I suspect that this can be written without `Parser.succeed identity`. Maybe just be removing it?

exercises/practice/sgf-parsing/.meta/src/SgfParsing.example.elm
```diff
@@ -0,0 +1,144 @@
...
+        |. Parser.symbol "("
+        |= Parser.lazy (\_ -> sgfWithoutParentheses)
+        |. Parser.symbol ")"
+
+
+sgfWithoutParentheses : Parser Tree
+sgfWithoutParentheses =
+    Parser.succeed (\prop trees -> Node { properties = prop, children = trees })
+        |. Parser.symbol ";"
+        |= Parser.loop [] properties
```
ceddlyburge, line 54: I think loop is less common than sequence, and the syntax of how it works looks a little strange to me. It feels a bit like an advanced topic.

It probably is better though with the tail call elimination, and other examples will be submitted over time anyway, so probably good to keep it as it is.

ceddlyburge (approved):
I've left some minor comments, some of which probably don't make sense, but this looks good to me.

I think my main question would be whether we can make this a concept exercise for a Parser concept?

I think we have concepts / concept exercises for all the prerequisites, and concept exercises are the things that we are short of (apart from some more automated analysis)

exercises/practice/sgf-parsing/.docs/instructions.md
```diff
@@ -0,0 +1,66 @@
...
+A key can have multiple values associated with it. For example:
+
+```text
+(;FF[4];AB[aa][ab][ba])
+```
+
+Here `AB` (add black) is used to add three black stones to the board.
+
+There are a few more complexities to SGF (and parsing in general), which
+you can mostly ignore. You should assume that the input is encoded in
```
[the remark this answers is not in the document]
jiegillet: I can't change the instructions, as they are centralized in the problem-specs repo.

exercises/practice/sgf-parsing/.docs/instructions.md
jiegillet: same as above.

exercises/practice/sgf-parsing/.meta/src/SgfParsing.example.elm
jiegillet: I had the same feeling as you and I tried when I was implementing it, but `(|.) : Parser keep -> Parser ignore -> Parser keep` needs something to keep on its left, so it can't be `Parser.symbol "("`.
I tried to find something better without success. I just now found that the docs use the same approach [in one of the examples](https://package.elm-lang.org/packages/elm/parser/latest/Parser#keyword) so I guess it's probably fine.

exercises/practice/sgf-parsing/.meta/src/SgfParsing.example.elm
jiegillet: I actually started with a sequence, and I think it worked, but all the properties (`start`, `stop`...) except for `item` were empty or irrelevant so it felt strange, especially using `spaces` felt like a waste of resources.
I agree with you that I would not use that for teaching about parsing, but this imoplementation is just meant as a proof that the exercise can be solved.

jiegillet:
> I think my main question would be whether we can make this a concept exercise for a Parser concept?

I think there is a policy against using practice exercises as concept exercises, so probably not. It's also a bit too hard. 
Once we have a parsing concept, this exercise should `practice` it for sure though.

> I think we have concepts / concept exercises for all the prerequisites, and concept exercises are the things that we are short of (apart from some more automated analysis)

I know we don't need more practice exercise, but it makes me so happy to add them haha :D
I might take this chance/momentum to work on a parsing concept though!

mpizenberg:
I remember that one of the exercices we designed for the exercism research
study had a parser approach. Maybe this can be adapted for a concept
exercise.

jiegillet:
Ah nice, I found them.
https://github.com/exercism/research_experiment_1/tree/main/exercises/elm-1-b
https://github.com/exercism/research_experiment_1/tree/main/exercises/elm-2-b
Yeah, that could help!

PR #502, opened 2022-09-14
Add practice exercise `circular-buffer`

jiegillet:
This one is all about implementing a data structure.

### DIFF
--- a/exercises/practice/circular-buffer/tests/Tests.elm
+++ b/exercises/practice/circular-buffer/tests/Tests.elm
@@ -1,0 +1,159 @@
+module Tests exposing (tests)
+
+import CircularBuffer
+import Expect
+import Test exposing (Test, describe, skip, test)
+
+
+type Operation a
+    = Read
+    | Write a
+    | Overwrite a
+    | Clear
+
+
+type Output a
+    = ReadSuccess a
+    | ReadFailure
+    | WriteSuccess
+    | WriteFailure
+
+
+run : Int -> List (Operation a) -> List (Output a)
+run size =
+    let
+        runOperation operation ( buffer, output ) =
+            case operation of
+                Read ->
+                    case CircularBuffer.read buffer of
+                        Just ( a, newBuffer ) ->
+                            ( newBuffer, ReadSuccess a :: output )
+
+                        Nothing ->
+                            ( buffer, ReadFailure :: output )
+
+                Write element ->
+                    case CircularBuffer.write element buffer of
+                        Just newBuffer ->
+                            ( newBuffer, WriteSuccess :: output )
+
+                        Nothing ->
+                            ( buffer, WriteFailure :: output )
+
+                Overwrite element ->
+                    ( CircularBuffer.overwrite element buffer, output )
+
+                Clear ->
+                    ( CircularBuffer.clear buffer, output )
+    in
+    List.foldl runOperation ( CircularBuffer.new size, [] )
+        >> Tuple.second
+        >> List.reverse
+
+
+tests : Test
+tests =
+    describe "CircularBuffer"
+        [ -- skip <|
+          test "reading empty buffer should fail" <|
+            \() ->
+                run 1 [ Read ]
+                    |> Expect.equal [ ReadFailure ]
+        , skip <|
+            test "can read an item just written" <|
+                \() ->
+                    run 1 [ Write 1, Read ]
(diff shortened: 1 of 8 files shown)


exercises/practice/circular-buffer/tests/Tests.elm
```diff
@@ -0,0 +1,159 @@
...
+                    run 2 [ Clear, Write 1, Write 2, Overwrite 3, Overwrite 4, Read, Read, Read ]
+                        |> Expect.equal
+                            [ WriteSuccess
+                            , WriteSuccess
+                            , ReadSuccess 3
+                            , ReadSuccess 4
+                            , ReadFailure
+                            ]
+        , skip <|
+            test "buffer can store strings" <|
```
jiegillet, line 150: This test doesn't exist in the canonical data, but one of the comments encourages tracks to add it for languages that can handle polymorphism. Seemed like worth doing.

exercises/practice/circular-buffer/tests/Tests.elm
```diff
@@ -0,0 +1,159 @@
...
+
+
+tests : Test
+tests =
+    describe "CircularBuffer"
+        [ -- skip <|
+          test "reading empty buffer should fail" <|
+            \() ->
+                run 1 [ Read ]
+                    |> Expect.equal [ ReadFailure ]
```
ceddlyburge, line 61: The assertion message when this test fails doesn't relate very well to the Circular Buffer problem domain, as ReadFailure is a test only thing.

And the same thing applies for all the tests I think.

Would asserting on the Circular Buffer itself make more sense?

ceddlyburge (approved):

exercises/practice/circular-buffer/tests/Tests.elm
jiegillet: I don't really want to do that, as it would mean imposing the Circular Buffer data structure, and coming up with the data structure is the whole point of the exercise to me.

I would argue that `ReadFailure` does relate to `read` since the signature is `read : CircularBuffer a -> Maybe ( a, CircularBuffer a )` that can succeed or fail, but I do agree that it is an unusual testing structure. I was trying to reproduce the structure of [the canonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/circular-buffer/canonical-data.json)
```
    {
      "uuid": "90741fe8-a448-45ce-be2b-de009a24c144",
      "description": "each item may only be read once",
      "property": "run",
      "input": {
        "capacity": 1,
        "operations": [
          {
            "operation": "write",
            "item": 1,
            "should_succeed": true
          },
          {
            "operation": "read",
            "should_succeed": true,
            "expected": 1
          },
          {
            "operation": "read",
            "should_succeed": false
          }
        ]
      },
      "expected": {}
    },
```
they insist that read and write should succeed or fail and that should be what is tested.
When making an implementation error, the error message is
```
↓ CircularBuffer
✗ overwrite acts like write on non-full buffer

    [WriteSuccess,ReadSuccess 2,ReadSuccess 1]
    ╷
    │ Expect.equal
    ╵
    [WriteSuccess,ReadSuccess 1,ReadSuccess 2]

```
and, as I was solving it myself, I thought it actually did the job of telling me what's wrong.

ceddlyburge: I see your point about the data structure, I hadn't noticed it wasn't part of the setup.

Read and write success is only one part of the equation though, you also need to know what was read / where was written to.

Would it be feasible to test an array / list like representation of the data, in the same way that it is described in the instructions?

jiegillet: I suppose it is possible, but it's not really how the tests are designed.
It's true that there is a gap between the instructions and the tests, but I'm having trouble coming up with something better. Would you have a concrete suggestion on how to achieve what you are proposing?

ceddlyburge: I don't want to be a pain, so feel free to merge if you think it is the right decision. You wrote this code so you know more about it than I do, so you are better placed to make the decision I think.

I don't have a plan for the code, but I thought the tests could start with a '[ ][ ][ ][1][2][3][ ]' type thing in the setup, maybe an array, may e a list, maybe a string, and then assert on the same structure. I think this would be better, but might not be worth the effort ...

PR #504, opened 2022-09-24
Add `introduction.md.tpl` files, unify introduction contents

jiegillet:
This PR is the equivalent of [this one](https://github.com/exercism/elixir/pull/1151) for Elixir.
The idea is to unify the concept exercise intro to the exercise intro, by adding `introduction.md.tpl` files and generating the intro via `configlet generate`.

Doing that, I have found several exercise that actually used the text from `about.md` instead of `introduction.md`, which is not recommended. I also found some typos that were not consistently fixed everywhere.

Another side effect of this is that a new header with the name of the topic is added (eg `## Tuples`) and all other heading levels are increased.

The reason I'm leaving this in draft is that
- I don't necessarily feel strongly about those changes, so I want to hear opinions about it
- There is a bug with configlet, it doesn't support slugs with numbers like our `basics-1` and `basics-2`, so I've left them alone for now. I've [opened an issue here](https://github.com/exercism/configlet/issues/663), but I wanted to have a branch that @ee7 can checkout for testing purposes.

### DIFF
--- a/exercises/concept/annalyns-infiltration/.docs/introduction.md
+++ b/exercises/concept/annalyns-infiltration/.docs/introduction.md
@@ -11,7 +11,7 @@ Booleans are also extremely useful to make branches in your code that only get e
 In Elm, this is done with an `if-then-else` expression, used as follows.
 
 ```elm
-if someCondition then
+if someCondition || (someOtherCondition && anotherCondition) then
     doSomething
 else
     doSomethingElse
--- a/exercises/concept/annalyns-infiltration/.docs/introduction.md
+++ b/exercises/concept/annalyns-infiltration/.docs/introduction.md
@@ -1,32 +1,13 @@
 # Introduction
 
+## Booleans
+
 The `Bool` type is one of the building blocks of the language.
 It aims at describing whether something is true or false, with the respective `True` and `False` values.
-The name `Bool` is short for Boolean, which is a field of mathematics related to logical thinking, named after the work of George Boole.
 
-## Boolean operators
+Booleans can be combined with the `&&` "AND" and the `||` "OR" operators, and negated with the `not` function.
 
-There are two functions and two operators to manipulate values of type `Bool`.
-
-- `not : Bool -> Bool`
-  This function transforms `True` into `False` and `False` into `True`.
-- `(&&) : Bool -> Bool -> Bool`
-  This operator called "AND" takes two booleans and returns `True` if they are both `True`.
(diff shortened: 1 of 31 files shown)


ceddlyburge:
This seems like a good idea to me, let me know if you want me to do anything ...

jiegillet:
OK great! I don't think you need to do anything.
I'll wait to hear on the configlet issue then, I don't want to commit code that will break things, and I would rather do this thing in batch.

jiegillet:
Thank you @ee7.
This is ready for review now.

exercises/concept/annalyns-infiltration/.docs/introduction.md
```diff
@@ -1,32 +1,13 @@
 # Introduction
 
+## Booleans
+
 The `Bool` type is one of the building blocks of the language.
 It aims at describing whether something is true or false, with the respective `True` and `False` values.
-The name `Bool` is short for Boolean, which is a field of mathematics related to logical thinking, named after the work of George Boole.
```
mpizenberg, line 5: I remember that at some point, there was a difference between the text visible in a concept page before starting the corresponding exercise and in the exercise introduction. So when it made sense, I tried to be more concise on the concept page and a bit more complete on the actual exercise introduction so that the student have enough relevant information for their exercise. I suppose that's why there is a difference in that concept exercise I wrote. Then I also added more info in the document available after finishing the concept (I don't remember the name of that doc).

Is this not the case anymore that there are two different purposes for these two documents? One just serves as a "concept teaser" and the other more complete for the actual exercise intro?

If those have been unified, then be careful about what you remove in this PR. It might remove useful info for students.

mpizenberg: Thanks for clearing that up for me. I also agree that it’s much easier to maintain the repo with the templates. As I don’t do much maintaining these day it’s probably better to align with your arguments. In that case, could you verify these intros that differ a lot and change the one used by the template to be enough introduction for students.

jiegillet:
@ceddlyburge @mpizenberg I managed to find time/motivation to review the introductions.
The let one was fine, I tweaked the boolean, maybe and tuple ones, mostly adding code examples, if you could check the last 3 commits, I would appreciate it.

jiegillet:
@ee7 works like a charm, thank you again!

mpizenberg (approved):
Thanks @jiegillet this is probably enough info now.

PR #503, opened 2022-09-23
Add concept `parsing` and concept exercise `paolas-prestigious-pizza`

jiegillet:
Closes #470.

I really enjoyed doing this one, looking forward to having your feedback.

PR #507, opened 2022-10-01
Add concept/concept exercise stub file generator

jiegillet:
This is a simple script that prepares all the necessary files for writing a new concept + concept exercise together.
I used it for preparing #503 and it was very helpful.
Obviously it includes no actual content, it's all stubs and placeholders apart from the concept and exercise names.

I was wondering if I should make it better by making actual template files to copy instead of doing `echo "content" > file` or also adding extra flags to the exercise generator so it would write the concept exercises to the right folder instead of `mv`ing them afterwards... 
But really, this does the job, so I though I would submit it this way.
What do you think?

### DIFF
--- a/bin/generate_concept_exercise.sh
+++ b/bin/generate_concept_exercise.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+# Exit if anything fails.
+set -euo pipefail
+
+# If argument not provided, print usage and exit
+if [[ $# -ne 2 ]]; then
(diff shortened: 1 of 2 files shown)


ceddlyburge (approved):
I'm not sure I would use this, I normally just copy and past an existing one, and configlet tells me if I forget the config.json.

I have no objections to it though, and it sounds like it works for your flow.

jiegillet:
Thank you, it definitely works for me, so I appreciate having it in the repo. 
Of course you don’t have to use it :)

Issue #429, opened 2021-09-19
Concept: composition


mpizenberg:
# Design

## Learning objectives

- Know how to compose functions of 1 argument with `<<`.
- Know how to compose functions of 1 argument with `>>`.
- Know how to compose partially applied functions with 1 argument left.
- Know about point-free style to define functions.
- Understand the limits of composition and point-free regarding readability.

## Out of scope

Nothing really out of scope.

## Concepts

- composition

## Prerequisites

- partial-application

PR #509, opened 2022-10-06
Partial application concept and concept exercise

### LINKED ISSUE #428: Concept: partial application
# Design

## Learning objectives

- Know that calling a function without all its arguments returns a new function.
- Understand the arrows notation `->` in type annotations.
- Know how convenient is partial applications in pipes `|>`.
- Understand how partial application and pipes encourage placement of the main data structure as last argument of functions.

## Out of scope

- Composition of partial functions is out of scope, left for another concept.

## Concepts

- partial-application

## Prerequisites

- basics-2

ceddlyburge:
Fixes #428


ceddlyburge:
@jiegillet @mpizenberg  can you take a look at this?

It's a new partial application concept and exercise. If you're happy with it I will also update the prerequisites of the other concept exercises that use the pipe operator

jiegillet:
Yay, new concept! Thank you!
I'll be away for the weekend (long weekend actually, Monday too), but I'll try to review it when I'm back next week.

ceddlyburge:
> I didn't like that I had to "pay attention to the order of the function parameters" but they were given in the wrong order, it felt like I was tricked. I think I would have preferred some explanation like "this function will be composed repeatedly to attack the monster later, think of the best configuration of the parameters (given in alphabetical order)"
I intuitively felt that the second function attackWithClaw would flip the order, but of course it didn't so I felt tricked again somehow. Now that I think about it, we might want to do that to show how much more hard it is to compose (???).

This is on purpose, so that you have to make a definite choice, and not just accept the default / obvious order of parameters. Maybe we could ask students to solve it twice, to highlight the difference?

> I don't like that the tests don't actually have a monster name and that the using an actual name would render "Giant SpiderAttacked with claw of strength 5." without a space and with a capital letter. The final attack string also looks bad.

We could add something for this, but I don't think it is necessary (certainly it isn't for the learning objective), and would add complication. Maybe we could call the alias `MonsterDamage` instead of Monster, with the implication that a Monster would have a MonsterDamage field?

> Maybe the sword could get damaged every time it's used so that its strength diminishes and we can justify having a parameter instead of hardcoding it (that's not a big deal though).

I'm not sure what the purpose of this is? It seems to add complication for no obvious learning benefit to me. Can you expand on what your thoughts are?

> It did feel good at the end when I had to compose the functions, it felt like it all made sense.

:) 

> I used the >> operator instead of |>, and I think it is a valid choice so it should also be introduced in this concept.

There is a composition concept, which I think is the better home for this. It could reuse this exercise though.

> No left pizza <|?

It is mentioned in the introduction, but it is a fairly rarely used thing in Elm, so no need to force a student to use one in the exercise I think. It's not mentioned as a goal in the issue either.

jiegillet:
> This is on purpose, so that you have to make a definite choice, and not just accept the default / obvious order of parameters. Maybe we could ask students to solve it twice, to highlight the difference?

I like the idea of making them solve both ways. The functions can be exactly the same, with parameters flipped.

> > I don't like that the tests don't actually have a monster name and that the using an actual name would render "Giant SpiderAttacked with claw of strength 5." without a space and with a capital letter. The final attack string also looks bad.
> 
> We could add something for this, but I don't think it is necessary (certainly it isn't for the learning objective), and would add complication. Maybe we could call the alias `MonsterDamage` instead of Monster, with the implication that a Monster would have a MonsterDamage field?

Yes, `MonsterDamage` or `MonsterHP` would be very easy to implement and quite intuitive to compose. 
```elm
attackWithSword : Int -> MonsterHP -> MonsterHP
attackWithSword strength hp = hp - strength
``` 

> 
> > Maybe the sword could get damaged every time it's used so that its strength diminishes and we can justify having a parameter instead of hardcoding it (that's not a big deal though).
> 
> I'm not sure what the purpose of this is? It seems to add complication for no obvious learning benefit to me. Can you expand on what your thoughts are?

Yes, you are right, this introduces complications for no clear benefits. Please disregard :)

> > I used the >> operator instead of |>, and I think it is a valid choice so it should also be introduced in this concept.
> 
> There is a composition concept, which I think is the better home for this. It could reuse this exercise though.

How would you feel about introducing 2 concepts with this exercise? The concepts are very close and work off each other in my opinion. We could add one function where `>>` is the best choice, maybe in a map `Something.map (doThis >> doThat)`.

> > No left pizza <|?
> 
> It is mentioned in the introduction, but it is a fairly rarely used thing in Elm, so no need to force a student to use one in the exercise I think. It's not mentioned as a goal in the issue either.

I was thinking they will almost certainly see it in the test files (weirdly almost the only place where `<|` is idiomatic) so they might be curious, it might be worth a quick mention (the docs do a great job in that aspect). I agree that we don't need to make them use it in the exercise.

ceddlyburge:
Hi @jiegillet , it sounds like we are approaching consensus!

I would be happy to include composition in the exercise, but I think we recently said that we should synchronise all the concept introductions and concept exercise introductions, so maybe we should merge the concepts.

So the concept would be "partial application and composition"

As the exercise demonstrates, they are very closely related, so I think it makes sense ...

jiegillet:
Sounds good to me!

ceddlyburge:
Fixes #429

ceddlyburge:
Hi @jiegillet , I've made those changes, can you take another look?
Thanks, Cedd

exercises/concept/monster-attack/.meta/design.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Prerequisites
+
+- basics-2
+
+## Analyzer
+
+This exercise could benefit from the following rules added to the the [analyzer][analyzer]:
+
+- Verify that the pipeline operator is used in `attack`
```
jiegillet, line 26: Update for attack1/2/3
The analyzer can't check the presence of operators yet, I need to add that (but it's simple, I did it for `let` recently)

jiegillet (commented):
Awesome!
I found a bunch of typos and stuff, so I think I would like to done one last pass after this if it's ok with you...

exercises/concept/monster-attack/.meta/design.md
```diff
@@ -0,0 +1,29 @@
...
+- Verify that attackWithClaw and attahWithSword are only used once each in `attack`
```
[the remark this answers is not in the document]
ceddlyburge: The idea is to use a partially applied function, like `annalynAttack = attackWithSword2 5`. You can take a look at the exemplar to see what I mean.

I think this would help a student learn partial application / currying, and it probably improves readability, although it is less concise. It's another example when currying can be useful. We could maybe add the name of the attacker (Annalyn / Kazak) in to the attack functions to make this more impactful, but it adds complexity as well, so not sure it is worth it.

ceddlyburge:
Hi @jiegillet , I've addressed those comments, can you take another look?
Thanks!
Cedd

jiegillet:
There are still a bunch if suggestions you missed. I'll merge them, as they are trivial.

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
+# Introduction
+
+## Partial application
+
+All functions in Elm are [curried][currying], which means that if you call a function without passing all of the arguments, it returns a new function.
+
+The result of doing this is called [partial-application][partial-application] (you are not passing all the arugments to the function, but are partially applying some of them).
+
+```elm
+add2 = add 2
```
jiegillet, line 10: we should add
```elm
add : Int -> Int -> Int
add a b = a + b
```

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+--> This is a function that takes a string, and returns True if that string contains Cedd)
+
+containsCedd "-Cedd-"
+--> True
+--> ("-Cedd-" ends up being the second argument to contains, and it does include "Cedd", the first argument)
+```
+
+## Pipe operator (`|>`)
+
+Saying `x |> f` is exactly the same as `f x`. A concrete example is `"5" |> String.toInt` is the same as `String.toInt "5"`
```
jiegillet, line 36:
```suggestion
Saying `x |> f` is exactly the same as `f x`. A concrete example is `"5" |> String.toInt` is the same as `String.toInt "5"`.
```

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+
+This doesn't sound very useful, but it is called the [pipe operator][pipe-operator] because it lets you write “pipelined” code.
```
jiegillet, line 38:
```suggestion
This doesn't sound very useful at first, but it is called the [pipe operator][pipe-operator] because it lets you write “pipelined” code.
```

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+    input
+    |> String.trim
+    |> String.toInt
+```
+
+This avoids the use of brackets, and shows the functions in the order that they are applied (first `String.trim`, then `String.toInt`), which aids readability.
+
+There is also a [reverse pipe operator `(<|)`][reverse-pipe-operator].
+Saying `x <| f` is exactly the same as `f x`.
+This doesn't look very useful at first glance, and is definitely not used as much as the pipe operator, but it still helps to avoid brackets in some situations.
```
jiegillet, line 62:
```suggestion
This doesn't look very useful at first glance either, and is definitely not used as much as the pipe operator, but it helps to avoid brackets in some situations.
```

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Function composition / point free style
+
+The [`(>>)`][forward-composition] operator creates a new function by composing two compatible functions together.
+It does this by taking the output from the first function and using this as the first argument to the second function.
+The created function will have the same parameters as the first function, and will return whatever the second function returns.
+The first function must return a type that the second function accepts as a parameter.
```
jiegillet, line 69: Should we add the types of the operator? It could help clarify, but it could also be more confusing? I'm giving it a shot, feel free to disregard.

```suggestion
The [`(>>)`][forward-composition] operator has the type `(a -> b) -> (b -> c) -> (a -> c)` and creates a function of type `a -> c`  by composing two compatible functions of type `a -> b `and `b -> c`.
It does this by taking the output (of type `b`) from the first function and using it as the first argument to the second function (which must be of the same type `b`).
The created function will take the same input (of type `a`) first function, and will return whatever `c` the second function returns.
```

concepts/partial-application-composition/introduction.md
```diff
@@ -0,0 +1,139 @@
...
+
+A concrete example is `String.trim >> String.toInt`.
+`String.trim` takes a string as its only parameter, so this is what the created function takes.
+`String.toInt` returns an `Int`, so this is what the created function returns
```
jiegillet, line 73:
```suggestion
`String.toInt` returns a `Maybe Int`, so this is what the created function returns.
```

exercises/concept/monster-attack/.docs/hints.md
```diff
@@ -0,0 +1,8 @@
+# General
+
+- [Partial application][partial-application] and [`(|>)`][pipe-operator] make it easy to write elegant, chained, code, when the main data structure as last argument of functions. All the common functions in the standard library do this, such as `List.map`
+- A [Let statement][let] can be used in `attack` to partially apply the `attackWithSword` and `attachWithClaw` functions with the correct strength for Annalyn and Kazak.
```
jiegillet, line 4:
```suggestion
- A [Let statement][let] can be used in the `attack` functions to partially apply the `attackWithSword` and `attachWithClaw` functions with the correct strength for Annalyn and Kazak.
```

exercises/concept/monster-attack/.docs/instructions.md
```diff
@@ -0,0 +1,62 @@
...
+
+## 1. Define attackWithSword1
+
+Define `attackWithSword1`.
+This should take a `MonsterDamage` (a `String` that describes the damage to the monster) and an `Int` that indicates how strong the attack is.
+It should return the existing string, with "Attacked with sword of strength <strength>." appended to the end.
+
+## 2. Define attackWithClaw1
+
+Define `attackWithClaw1`. This should take a `MonsterDamage` and an `Int` that indicates how strong the attack is.
```
jiegillet, line 15:
```suggestion
Define `attackWithClaw1`.
This should take a `MonsterDamage` and an `Int` that indicates how strong the attack is.
```

jiegillet (approved):
OK, nice!
I went through it again and left a few more typos and suggestions, I leave them up to you.

PR #517, opened 2022-10-29
Configlet sync: tests

jiegillet:
Synced all new tests. 
Most were straightforward, some revealed a bug in implementations (I added my name to contributors after fixing it), some were not implemented (I left justifications in tests.toml).


mpizenberg (approved):
Thanks!

PR #520, opened 2022-10-30
Add new practice exercise: `custom-set`

jiegillet:
This is my way of getting hyped to finish the `set` concept: totally procrastinating and implementing my own set via a practice exercise (well, we can use this one to practice sets, even though using `Set` is prohibited).

This one involves creating a data structure, which I love.

I will need an analyzer rule for checking that students don't use `Set` at all.

### DIFF
--- a/exercises/practice/custom-set/.meta/src/CustomSet.example.elm
+++ b/exercises/practice/custom-set/.meta/src/CustomSet.example.elm
@@ -1,0 +1,198 @@
+module CustomSet exposing (diff, disjoint, empty, equal, fromList, insert, intersect, isEmpty, member, subset, toList, union)
+
+
+type Set
+    = Nil
+    | Node Int Set Set
+
+
+empty : Set
+empty =
+    Nil
+
+
+insert : Int -> Set -> Set
+insert element set =
+    case set of
+        Nil ->
+            Node element Nil Nil
+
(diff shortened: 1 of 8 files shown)


exercises/practice/custom-set/.meta/src/CustomSet.example.elm
```diff
@@ -0,0 +1,198 @@
+module CustomSet exposing (diff, disjoint, empty, equal, fromList, insert, intersect, isEmpty, member, subset, toList, union)
```
ceddlyburge, line 1: This seems like a complicated example! It seems like there would be a simpler solution. Not sure if that matters or not.

ceddlyburge (approved):

exercises/practice/custom-set/.meta/src/CustomSet.example.elm
jiegillet: Of course, the instructions say that anything is fine, so it would be easier to model a set as a list, but where's the fun in that? 😆 
I was actually thinking that my solution was too simple because my tree is unbalanced, but implementing a red-black tree might be a bit much.

Ultimately, like you are saying, it doesn't matter, this is not a reference solution, just a proof that it can be solved in Elm.

PR #521, opened 2022-10-30
Update title of partial composition concept

### LINKED ISSUE #519: Incorrectly formatted partial application concept header
[The concept page for partial application composition](https://exercism.org/tracks/elm/concepts/partial-application-composition) is showing partial-application-composition as the concept heading. Presumably, it should be Partial Application Composition judging from the other concept pages.

ceddlyburge:
Fixes #519

### DIFF
--- a/config.json
+++ b/config.json
@@ -133,7 +133,7 @@
     },
     {
       "slug": "partial-application-composition",
-      "name": "partial-application-composition",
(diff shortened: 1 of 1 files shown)


mpizenberg (approved):
Thanks for the fix!

PR #524, opened 2022-11-03
Add icons to concept exercises

jiegillet:
We had several exercises without icons, I found some that were suitable [from here](https://github.com/exercism/website-icons/blob/main/exercises/README.md).
Let me know if you have better suggestions.


jiegillet:
I think it's as simple as opening an issue on the `website-icons` repo, but it usually takes a really long time, so if another icon fits the bill, might as well../

PR #525, opened 2022-11-04
Review test cases not included

### LINKED ISSUE #516: exercises: practice: tests: consider implementing some `include = false` tests
PR https://github.com/exercism/elm/pull/490/ synced tests, but excluded every new test. Should these be included? If so, this issue can track them.

jiegillet:
Closes #516.
@ee7 noticed that a PR left tests cases unimplemented, so I reviewed all these cases.

Most were reimplemented by other tests, some tests were already there, one was not included for valid reasons (I added a comment), some were easy to include and one required a simple change in the example solution (I added myself to the contributor list).

@ee7 while going through this, I was thinking that it would be nice for configlet to add a comment to tests reimplemented by other tests (like there is a comment tests that reimlement others).


ceddlyburge (approved):
I was wondering what that issue was about, now I see it is to do with include=false in the various tests.toml files.

PR #527, opened 2022-11-12
Run elm-test-runner in the CI

### LINKED ISSUE #412: Add CI to prevent usage of Fuzz testing
Fuzz testing is great, but this is not handled yet by the code in the test runner that extracts the test code to show it to a user of the online editor. So we should make sure that Fuzz testing is not introduced unintentionally.

jiegillet:
Closes #412.

~~This will have to wait for [this PR](https://github.com/exercism/elm-test-runner/pull/43) to be merged before working, so I'm opening it as a draft for reference.~~
https://github.com/exercism/elm-test-runner/pull/43 was dropped, see justifications over there.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,4 +46,4 @@ jobs:
           path: elm_repo
 
       - name: Run tests for all exercises
-        run: ./bin/run-all-exercises-in-docker.sh elm_repo
+        run: elm_repo/bin/run-all-exercises-in-docker.sh elm_repo
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout PR
-        uses: actions/checkout@v3
+        uses: actions/checkout@e2f20e631ae6d7dd3b768f56a5d2af784dd54791
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3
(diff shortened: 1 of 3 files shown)


jiegillet:
This is ready now. The scripts are identical (up to path differences) to https://github.com/exercism/elm-test-runner/pull/43 which was already reviewed and approved, the only different thing here is  that it runs in the CI, that runs fine.
@mpizenberg could you have a look?

ceddlyburge:
> This will have to wait for [this PR](https://github.com/exercism/elm-test-runner/pull/43) to be merged before working, so I'm opening it as a draft for reference.

Tihs is no longer the case I think?

ceddlyburge:
This looks good to me, although there is a lot of "tinyproxy: Could not create listening sockets." in the output, and it might be good to show the passing tests in the output as well, to aid confidence / observability that everything really was being tested.

I also note that you are using the elm-test-runner to do this, when it seems that you could use elm-test or elm-test-rs directly. Was there a reason for this approach? Might be good to document it.

I would be happy to approve this, but i think you want Matthieu to do so.

mpizenberg (approved):
Do we actually need to clone and build the docker of the runner considering it already exists pre-built for exercism? (I think both in aws and docker? @ErikSchierboom).
We might want to actually rebuild it specifically if we are changing the elm-test-runner repo and want to check everything before pushing images. In that case, we could also cache the builds in the CI.
Anyway, all this can be discussed in a follow up PR or left as is, since the current 2 min run is very reasonable.

mpizenberg:
> I also note that you are using the elm-test-runner to do this, when it seems that you could use elm-test or elm-test-rs directly. Was there a reason for this approach? Might be good to document it.

Yep, the goal was to check that the docker image of the test runner works as expected, as it’s the one running for students on the website.

> it might be good to show the passing tests in the output as well, to aid confidence / observability that everything really was being tested

That’s a good idea

ErikSchierboom:
> Do we actually need to clone and build the docker of the runner considering it already exists pre-built for exercism? (I think both in aws and docker? @ErikSchierboom).

Not necessarily, you could do `docker pull exercism/elm-test-runner` and get the version that's currently deployed to production.

jiegillet:
Nice, `docker pull` removed 30 seconds to the run, it's under a minute now.
I also added some logging. 

>  We might want to actually rebuild it specifically if we are changing the elm-test-runner repo and want to check everything before pushing images. In that case, we could also cache the builds in the CI.

That sounds like something that should live in the elm-test-runner repo then.

Thank you @ErikSchierboom  and @mpizenberg, I learned a lot with this PR.

PR #532, opened 2022-11-17
Add generics concept and concept exercise

### LINKED ISSUE #422: Concept: generics
# Design

## Goal

The goal of this exercise is to teach students how to consume and make generic code.

## Learning objectives

- Understand what a type variable is.
- Know how to define generic types.
- Know how to define generic functions.

## Out of scope

- Covariance and contravariance.
- Higher order functions.

## Concepts

- `generics`: Learn how to create and use Elm code that is generic.

## Prerequisites

- `custom-types`: Learn how to use custom types in an Elm program.

ceddlyburge:
Fixes #422 

The exercise requires the use of a generic type parameter in a custom type and in a function type signature, but not in a record.

We could potentially ask the student to solve the exercise using a record first, and then using a custom type, similar to how we do it in the Monster Attack partial application exercise. And then in the opaque type concept / exercise we go one step further, making it impossible to get the treasure without the password. It might be a nice journey from simpler and easier to more complicated bit more robust. 

Or we could take the view that the way generics are used in records and custom types are similar enough that we don't need the student to explicitly code up both variants, and can see how it is done from the about / introduction.

@jiegillet @mpizenberg

### DIFF
--- a/concepts/generics/about.md
+++ b/concepts/generics/about.md
@@ -1,0 +1,77 @@
+# About
+
+# Introduction
+
+Generics are a language feature that enable code reuse.
+For example, it should not matter what type a list holds in order to compute its length.
+That is why the `length` function uses a generic type parameter `a` in its type signature.
+
+```elm
+length : List a -> Int
+```
+
+## Parameters in types
+
+When a type is generic over another type, we use a lowercase name for that other type, and call it a type parameter.
+Two very common examples are the `List` and `Maybe` types.
+
+```elm
+-- Maybe is generic over the type a
+type Maybe a
+    = Nothing
+    | Just a
+
+someString : Maybe String
+someString = Just "hello"
+
+someInt : Maybe Int
+someInt = Just 42
+```
+
+A type can be generic over multiple other types, and thus have multiple type parameters.
+In that case, every independant type gets a different lowercase name.
+Being independant does not prevent them from becoming the same type, it just enables them to be different.
+
+```elm
+type alias Pair a b =
+    { first : a
+    , second : b
+    }
+
+person : Pair String Int
+person = { first = "Jane", second = 18 }
+
+coordinates : Pair Int Int
+coordinates = { first = -3, second = 15 }
+```
+
+One important remark about names of type parameters, is that they can be multiple characters long.
+The only constraint is that they must be lowercase names.
+As such, when useful, we tend to give them descriptive names.
+
+> This is why the `Html` package defines the `Html msg` type, with one type parameter named `msg`.
+> The elm architecture guides you towards creation of a custom type `Msg` describing all messages that your web page can trigger.
+> As a result, your code will be building views of the type `Html Msg` where `Msg` is the concrete type that you defined.
+> Beginners are sometimes confused between errors mentioning `Html msg` (lowercase) and errors with `Html Msg`.
+> Now you know!
+> The difference is that the first one is generic over any `msg` type, while the second is specific to your `Msg` type.
+
+## Type parameters in functions
+
+Since types can be generic, so can functions.
+Functions handling generic types can be writen, such as the `List.length` function that we mentioned at the beginning.
+
+```elm
+length : List a -> Int
+length list =
+    case list of
+        [] ->
+            0
+        _ :: restOfTheList ->
+            1 + length restOfTheList
+```
+
+Type parameters can appear both in the inputs and in the output of a function, and genericity enabled by type parameters is usually called **parametric polymorphism**.
+This is in contrast with ad hoc polymorphism also known as function overloading, where the same function name can be used on multiple implementation where the types differ.
+This also contrasts with subtype polymorphism, where types can be defined in a hierarchy, and functions defined on a type can also be called with arguments of its subtypes.
+In elm, the only available mechanism for genericity is parametric polymorphism.
--- a/concepts/generics/links.json
+++ b/concepts/generics/links.json
@@ -1,0 +1,6 @@
+[
+  {
+    "url": "https://elmprogramming.com/type-system.html",
+    "description": "Elm type system"
+  }
+]
--- a/config.json
+++ b/config.json
@@ -19,18 +19,10 @@
     "analyzer": true
   },
   "files": {
-    "solution": [
-      "src/%{pascal_slug}.elm"
-    ],
-    "test": [
-      "tests/Tests.elm"
-    ],
-    "example": [
-      ".meta/src/%{pascal_slug}.example.elm"
-    ],
-    "exemplar": [
-      ".meta/Exemplar.elm"
-    ]
+    "solution": ["src/%{pascal_slug}.elm"],
+    "test": ["tests/Tests.elm"],
+    "example": [".meta/src/%{pascal_slug}.example.elm"],
+    "exemplar": [".meta/Exemplar.elm"]
   },
   "tags": [
     "paradigm/declarative",
@@ -154,6 +146,11 @@
       "slug": "set",
       "name": "Set",
       "uuid": "cfa44843-b783-4366-a198-f0e99e632143"
+    },
+    {
+      "slug": "generics",
+      "name": "Generics",
+      "uuid": "a967b004-e754-4664-8103-6a98be4d00e5"
     }
   ],
   "exercises": {
@@ -163,9 +160,7 @@
         "slug": "lucians-luscious-lasagna",
         "name": "Lucian's Luscious Lasagna",
         "uuid": "54b70493-702b-4577-9d47-ec5ede4cc2c1",
-        "concepts": [
-          "basics-1"
-        ],
+        "concepts": ["basics-1"],
         "prerequisites": [],
         "status": "beta"
       },
@@ -173,94 +168,63 @@
         "slug": "bettys-bike-shop",
         "name": "Betty's Bike Shop",
         "uuid": "88818c16-fff8-4964-b9a4-c799833ff970",
-        "concepts": [
-          "basics-2"
-        ],
-        "prerequisites": [
-          "basics-1"
-        ],
+        "concepts": ["basics-2"],
+        "prerequisites": ["basics-1"],
         "status": "beta"
       },
       {
         "slug": "annalyns-infiltration",
         "name": "Annalyn's Infiltration",
         "uuid": "9f501a73-14e3-4ada-8359-fe051530554a",
-        "concepts": [
-          "booleans"
-        ],
-        "prerequisites": [
-          "basics-2"
-        ],
+        "concepts": ["booleans"],
+        "prerequisites": ["basics-2"],
         "status": "beta"
       },
       {
         "slug": "tracks-on-tracks-on-tracks",
         "name": "Tracks on Tracks on Tracks",
         "uuid": "07b1a020-d534-4449-8698-2006e1176b3a",
-        "concepts": [
-          "lists"
-        ],
-        "prerequisites": [
-          "basics-2"
-        ],
+        "concepts": ["lists"],
+        "prerequisites": ["basics-2"],
         "status": "beta"
       },
       {
         "slug": "bandwagoner",
         "name": "Bandwagoner",
         "uuid": "af1cc0cd-b9f8-46ef-b4bf-7a5a17b1c8d6",
-        "concepts": [
-          "records"
-        ],
-        "prerequisites": [
-          "basics-2"
-        ],
+        "concepts": ["records"],
+        "prerequisites": ["basics-2"],
         "status": "beta"
       },
       {
         "slug": "valentines-day",
         "name": "Valentines Day",
         "uuid": "2ce85ab7-35a0-4566-803f-e60397a2391d",
-        "concepts": [
-          "custom-types"
-        ],
-        "prerequisites": [
-          "basics-2"
-        ],
+        "concepts": ["custom-types"],
+        "prerequisites": ["basics-2"],
         "status": "beta"
       },
       {
         "slug": "tisbury-treasure-hunt",
         "name": "Tisbury Treasure Hunt",
         "uuid": "d88d7e37-f52b-4999-b7df-bf0db09e7ac5",
-        "concepts": [
-          "tuples"
-        ],
-        "prerequisites": [
-          "lists",
-          "partial-application-composition"
-        ],
+        "concepts": ["tuples"],
+        "prerequisites": ["lists", "partial-application-composition"],
         "status": "beta"
       },
       {
         "slug": "role-playing-game",
         "name": "Role Playing Game",
         "uuid": "c62c5128-5733-4187-b46a-80660d9b5725",
-        "concepts": [
-          "maybe"
-        ],
-        "prerequisites": [
-          "records"
-        ],
+        "concepts": ["maybe"],
+        "prerequisites": ["records"],
         "status": "beta"
       },
       {
         "slug": "go",
         "name": "Go",
         "uuid": "33c34af3-26fa-4e18-8bb3-8af688069c9e",
-        "concepts": [
-          "result"
-        ],
+        "concepts": ["result"],
         "prerequisites": [
           "maybe",
           "basics-2",
@@ -275,9 +239,7 @@
         "slug": "ticket-please",
         "name": "Ticket, Please!",
         "uuid": "58f70123-5e59-4bac-b1e0-3b8504de360b",
-        "concepts": [
-          "pattern-matching"
-        ],
+        "concepts": ["pattern-matching"],
         "prerequisites": [
           "maybe",
           "records",
@@ -292,21 +254,15 @@
         "slug": "marios-marvellous-lasagna",
         "name": "Mario's marvellous lasagna",
         "uuid": "8b5b43a1-79ec-4f2f-a8ac-8aa2955a4674",
-        "concepts": [
-          "let"
-        ],
-        "prerequisites": [
-          "basics-2"
-        ],
+        "concepts": ["let"],
+        "prerequisites": ["basics-2"],
         "status": "beta"
       },
       {
         "slug": "top-scorers",
         "name": "Top Scorers",
         "uuid": "67b35178-0a74-4265-84a7-276c018cc3a5",
-        "concepts": [
-          "dict"
-        ],
+        "concepts": ["dict"],
         "prerequisites": [
           "booleans",
           "maybe",
@@ -319,9 +275,7 @@
         "slug": "paolas-prestigious-pizza",
         "name": "Paola's Prestigious Pizza",
         "uuid": "9ecdb4fc-d175-4058-9167-948b2a88fa1b",
-        "concepts": [
-          "parsing"
-        ],
+        "concepts": ["parsing"],
         "prerequisites": [
           "basics-1",
           "basics-2",
@@ -337,22 +291,14 @@
         "slug": "monster-attack",
         "name": "Monster Attack",
         "uuid": "c8abe3c6-44f8-49cd-b8e4-9951c77ddbd9",
-        "concepts": [
-          "partial-application-composition"
-        ],
-        "prerequisites": [
-          "basics-1",
-          "basics-2",
-          "let"
-        ]
+        "concepts": ["partial-application-composition"],
+        "prerequisites": ["basics-1", "basics-2", "let"]
       },
       {
         "slug": "blorkemon-cards",
         "name": "Blorkemon Cards",
         "uuid": "90660cf2-9247-41c0-8849-eee3e7118371",
-        "concepts": [
-          "comparison"
-        ],
+        "concepts": ["comparison"],
         "prerequisites": [
           "booleans",
           "tuples",
@@ -366,15 +312,16 @@
         "slug": "gotta-snatch-em-all",
         "name": "Gotta Snatch'Em All",
         "uuid": "a019f08f-b20e-4430-b01a-26b25a2bd0bb",
-        "concepts": [
-          "set"
-        ],
-        "prerequisites": [
-          "booleans",
-          "tuples",
-          "lists",
-          "comparison"
-        ],
+        "concepts": ["set"],
+        "prerequisites": ["booleans", "tuples", "lists", "comparison"],
+        "status": "beta"
+      },
+      {
+        "slug": "treasure-chest",
+        "name": "Treasure Chest",
+        "uuid": "dbfbca75-01c6-487e-8bb7-22854453489e",
+        "concepts": ["generics"],
+        "prerequisites": ["custom-types", "maybe", "pattern-matching"],
         "status": "beta"
       }
     ],
@@ -386,9 +333,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "strings"
-        ]
+        "topics": ["strings"]
       },
       {
         "slug": "two-fer",
@@ -397,9 +342,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "strings"
-        ]
+        "topics": ["strings"]
       },
       {
         "slug": "bob",
@@ -408,11 +351,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "conditionals",
-          "strings",
-          "text_formatting"
-        ]
+        "topics": ["conditionals", "strings", "text_formatting"]
       },
       {
         "slug": "leap",
@@ -421,10 +360,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "conditionals",
-          "integers"
-        ]
+        "topics": ["conditionals", "integers"]
       },
       {
         "slug": "space-age",
@@ -433,9 +369,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "floating_point_numbers"
-        ]
+        "topics": ["floating_point_numbers"]
       },
       {
         "slug": "sum-of-multiples",
@@ -444,11 +378,7 @@
         "practices": [],
         "prerequisites": [],
         "difficulty": 1,
-        "topics": [
-          "lists",
-          "math",
-          "transforming"
-        ]
+        "topics": ["lists", "math", "transforming"]
       },
       {
         "slug": "accumulate",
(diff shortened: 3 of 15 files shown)


concepts/generics/about.md
```diff
@@ -0,0 +1,77 @@
...
+someString : Maybe String
+someString = Just "hello"
+
+someInt : Maybe Int
+someInt = Just 42
+```
+
+A type can be generic over multiple other types, and thus have multiple type parameters.
+In that case, every independant type gets a different lowercase name.
+Being independant does not prevent them from becoming the same type, it just enables them to be different.
```
jiegillet, line 33:
```suggestion
In that case, every independent type gets a different lowercase name.
Being independent does not prevent them from becoming the same type, it just enables them to be different.
```

concepts/generics/about.md
```diff
@@ -0,0 +1,77 @@
...
+coordinates : Pair Int Int
+coordinates = { first = -3, second = 15 }
+```
+
+One important remark about names of type parameters, is that they can be multiple characters long.
+The only constraint is that they must be lowercase names.
+As such, when useful, we tend to give them descriptive names.
+
+> This is why the `Html` package defines the `Html msg` type, with one type parameter named `msg`.
+> The elm architecture guides you towards creation of a custom type `Msg` describing all messages that your web page can trigger.
```
jiegillet, line 53:
```suggestion
> The Elm architecture guides you towards creation of a custom type `Msg` describing all messages that your web page can trigger.
```

concepts/generics/about.md
```diff
@@ -0,0 +1,77 @@
...
+        [] ->
+            0
+        _ :: restOfTheList ->
+            1 + length restOfTheList
+```
+
+Type parameters can appear both in the inputs and in the output of a function, and genericity enabled by type parameters is usually called **parametric polymorphism**.
+This is in contrast with ad hoc polymorphism also known as function overloading, where the same function name can be used on multiple implementation where the types differ.
+This also contrasts with subtype polymorphism, where types can be defined in a hierarchy, and functions defined on a type can also be called with arguments of its subtypes.
+In elm, the only available mechanism for genericity is parametric polymorphism.
```
jiegillet, line 77:
```suggestion
In Elm, the only available mechanism for genericity is parametric polymorphism.
```

concepts/generics/about.md
```diff
@@ -0,0 +1,77 @@
...
+> The elm architecture guides you towards creation of a custom type `Msg` describing all messages that your web page can trigger.
+> As a result, your code will be building views of the type `Html Msg` where `Msg` is the concrete type that you defined.
+> Beginners are sometimes confused between errors mentioning `Html msg` (lowercase) and errors with `Html Msg`.
+> Now you know!
+> The difference is that the first one is generic over any `msg` type, while the second is specific to your `Msg` type.
+
+## Type parameters in functions
+
+Since types can be generic, so can functions.
+Functions handling generic types can be writen, such as the `List.length` function that we mentioned at the beginning.
```
jiegillet, line 62: Let's add `List.map` as an example too to show generics directly in functions

```elm
map : (a -> b) -> List a -> List b
map function list =
    case list of
        [] ->
            []
        head :: restOfTheList ->
            function head :: map function restOfTheList
```

concepts/generics/about.md
```diff
@@ -0,0 +1,77 @@
...
+someInt : Maybe Int
+someInt = Just 42
+```
+
+A type can be generic over multiple other types, and thus have multiple type parameters.
+In that case, every independant type gets a different lowercase name.
+Being independant does not prevent them from becoming the same type, it just enables them to be different.
+
+```elm
+type alias Pair a b =
```
jiegillet, line 36: What do you think of showing an example of a sum type on top of the product type?

```elm
type Either a b
  = This a
  | That b

this : Either a b -> Maybe a
this either =
  case either of
    This a -> Just a
    That _ -> Nothing
```

If it's too much, maybe it can be in about.md but not in introduction.md?

concepts/generics/links.json
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "url": "https://elmprogramming.com/type-system.html",
+    "description": "Elm type system"
+  }
```
jiegillet, line 5: We can add the ones from the hints.

exercises/concept/treasure-chest/.docs/instructions.md
```diff
@@ -0,0 +1,26 @@
+# Instructions
+
+In this exercise you're going to write a generic (/ magical!) TreasureChest, to store some treasure.
+
+## 1. Define the TreasureChest generic custom type
+
+Define a `TreasureChest` custom type with a single variant.
+
+The variant should have an associated String value, which will be used to store the password of the treasure chest
```
jiegillet, line 9:
```suggestion
The variant should have an associated `String` value, which will be used to store the password of the treasure chest.
```

exercises/concept/treasure-chest/.docs/instructions.md
```diff
@@ -0,0 +1,26 @@
...
+
+The variant should also have an associate generic value, which will be used to store the treasure. This value is generic so that the treasure can be anything.
```
jiegillet, line 11:
```suggestion
The variant should also have an associate generic value, which will be used to store the treasure.
This value is generic so that the treasure can be anything.
```

exercises/concept/treasure-chest/.docs/instructions.md
```diff
@@ -0,0 +1,26 @@
...
+
+## 2. Define the getTreasure function
+
+This function should take two parameters
+
+- a String (for the passwordAttempt)
+- a TreasureChest generic custom type
```
jiegillet, line 18:
```suggestion
- a `String` (for trying a password)
- a `TreasureChest` generic custom type
```

exercises/concept/treasure-chest/.docs/instructions.md
```diff
@@ -0,0 +1,26 @@
...
+
+This function should check the provided passwordAttempt against the String in the TreasureChest.
+
+The function should return a `Maybe`
+
+If the passwords match then return `Just` with the generic value from the TreasureChest (the treasure!)
```
jiegillet, line 24:
```suggestion
This function should check the provided password attempt against the `String` in the `TreasureChest`.

The function should return a `Maybe`.

If the passwords match then return `Just` with the generic value from the `TreasureChest` (the treasure!)
```

exercises/concept/treasure-chest/.meta/design.md
```diff
@@ -0,0 +1,26 @@
...
+## Learning objectives
+
+- Understand what a type variable is.
+- Know how to define generic types.
+- Know how to define generic functions.
+
+## Out of scope
+
+- Covariance and contravariance.
+- Higher order functions.
```
jiegillet, line 16: I think we have the bandwidth for talking about higher order functions, we can add one paragraph with `List.map` like I suggested before.

exercises/concept/treasure-chest/.meta/design.md
```diff
@@ -0,0 +1,26 @@
+# Design
```
jiegillet, line 1: Currently there is no need for an analyzer rule that I can think of, but let's add the category to make it explicit.

exercises/concept/treasure-chest/tests/Tests.elm
```diff
@@ -0,0 +1,33 @@
...
+
+import Expect
+import Test exposing (..)
+import TreasureChest exposing (..)
+
+
+tests : Test
+tests =
+    describe "TreasureChest"
+        [ describe "1"
```
jiegillet, line 11: The `"1"` tests should pass when the conditions are cleared for the first task, but currently they do not, because the code won't compile without `getTreasure` defined.
I suggest to add it naked, without a signature or arguments:
```elm
getTreasure =
  Debug.todo """
define getTreasure, with two parameters:
- a String (for the passwordAttempt)
- a TreasureChest custom type
 and returns:
- if the passwordAttempt does not match the String in the TreasureChest
  - Nothing
- if the passwordAttempt matches the String in the TreasureChest
  - Just treasure (where treasure is the generic in the TreasureChest)
"""
```
(but either simplify the text or make it match exactly the instructions once they are finalized)

jiegillet (commented):
Nice work!
At first, I liked the idea of asking the student to use a record and then a custom type, but I changed my mind. One problem is that we have no way of checking what they used with tests, and also there is no better style in general, it really depends on the system you are trying to model. But I like the idea of going further in follow-up concepts.

As always, I first tried to solve the exercise just by looking at the instructions and I came up with
```elm
module TreasureChest exposing (..)


type TreasureChest a
    = TreasureChest String a


getTreasure : String -> TreasureChest a -> Maybe a
getTreasure password (TreasureChest truePassword treasure) =
    if password == truePassword then
        Just treasure

    else
        Nothing
```
which is exactly your solution (up to normalization) 🎉 .

It's quite nice, but I would like to go a little deeper with generics and introduce higher order functions. I was thinking of asking them to define `swapTreasure : (a -> b) -> Treasure a -> Treasure b`, (think Indiana Jones swapping the gold statue for a bag of sand) or `swapTreasure : (a -> b) -> String -> Treasure a -> Maybe (Treasure b)` if you want to use the password.
It's not that much harder, but it might make them think about types more.

exercises/concept/treasure-chest/.docs/hints.md
```diff
@@ -0,0 +1,16 @@
+# General
+
+## 1. Define the TreasureChest generic custom type
+
+- The Elm guide has a section on [custom types][custom-types].
+- The TreasueChest custom type only has a single variant.
+- Idiomatically in Elm, when a custom type has a single variant, the variant and the type have the same name (in this case both would be called TreasureChest)
```
jiegillet, line 7:
```suggestion
- The `TreasureChest` custom type only has a single variant.
- Idiomatically in Elm, when a custom type has a single variant, the variant and the type have the same name (in this case both would be called `TreasureChest`)
```

exercises/concept/treasure-chest/.docs/hints.md
```diff
@@ -0,0 +1,16 @@
...
+
+## 2. Define the getTreasure function
+
+- There is a [shorthand for destructuring single member custom types][single-member-custom-destructuring].
+- The Elm guide has a sectoin on [Maybe][maybe] if you need a refresher.
```
jiegillet, line 12:
```suggestion
- The Elm guide has a section on [Maybe][maybe] if you need a refresher.
```

jiegillet:
Oh we should look for practice exercises that practice this concept. From the top of my head `circular-buffer` is a good fit, there might be more.

ceddlyburge:
> Oh we should look for practice exercises that practice this concept. From the top of my head `circular-buffer` is a good fit, there might be more.

I think I have added some of these in the config.json, but I guess the formatting changes have obscured them. I must have a plugin that is autoformatting these, which seems like a sensible idea, but we all need to use the same one / same settings for it to work. Do you have any thoughts about this? I was thinking that configlet fmt would be responsible for this, but it doesn't seem like it is.

ceddlyburge:
> It's quite nice, but I would like to go a little deeper with generics and introduce higher order functions. I was thinking of asking them to define swapTreasure : (a -> b) -> Treasure a -> Treasure b, (think Indiana Jones swapping the gold statue for a bag of sand) 

I like the Indiana Jones angle!

I think the swapTreasure function would be more idiomatically written as TreasureChest.map though. And I think the function would obey the functor laws, and thus make TreasureChest a functor. So I think I would prefer to teach this in a functor concept. That concept and exercise would follow on from this one quite nicely though.

exercises/concept/treasure-chest/.meta/design.md
ceddlyburge: I've discussed this in answer to another comment ...

ceddlyburge:
Hi @jiegillet, I've responded to all of those comments, and done most of them.
I'm open to ideas about what to use to format the json files, what do you use?
And the remaining point to iron out is the swapTreasure / functor issue.
Cheers, Cedd

exercises/concept/treasure-chest/.docs/instructions.md
jiegillet: @ceddlyburge you missed some of the comments, I think you didn't click "load more" to see them all.

jiegillet:
> I think I have added some of these in the config.json, but I guess the formatting changes have obscured them. I must have a plugin that is autoformatting these, which seems like a sensible idea, but we all need to use the same one / same settings for it to work. Do you have any thoughts about this? I was thinking that configlet fmt would be responsible for this, but it doesn't seem like it is.

Indeed, I didn't notice at all since all lists were reformatted.
`configlet fmt` doesn't do formatting, I think it just checks the fields. I use VSCode and I believe its default json formatter happens to be the right one. Passing the file through `jq '.'` also formats it correctly with one exception: the `2.0` in`"test_runner": { "average_run_time": 2.0 }` always gets outputted as `2` but `configlet fmt` requires a float. When I use the exercise generator this always happens, so it's a bit of a pain. In the Elixir track we changed the values to `5.1` instead of `5.0` but that just looks weird :)
Anyway, I think the main to settings are:
- Expand all objects and lists (that's the one you don't have)
- 2 space indentation

> I think the swapTreasure function would be more idiomatically written as TreasureChest.map though. And I think the function would obey the functor laws, and thus make TreasureChest a functor. So I think I would prefer to teach this in a functor concept. That concept and exercise would follow on from this one quite nicely though.

I don't really mind calling it `map`, but I think we should **not** have a functor concept in the track. For two reasons:
- Category theory nomenclature is not at all a part of the Elm culture. Of course the people who are curious learn about it eventually, but the docs never mention it and never rely on it.
- It's not entirely clear what is a functor and what isn't in Elm. In Haskell, `Set a` is not a functor, even if it has a `map` defined, because of the dependence on `Ord a` ([link1](https://www.reddit.com/r/haskell/comments/2090x3/ask_rhaskell_why_is_there_no_functor_instance_for/), [link2](https://www.schoolofhaskell.com/user/snoyberg/random-code-snippets/set-is-not-a-functor)), basically because Haskell accepts endofunctors in the `Hask` category. In Elm, it's simply not a valid question.

For these reasons, I still think `swapTreasure` or `TreasureChest.map` is a good fit here, because this exercise is about implementing a data structure and mapping is a very natural concept.

exercises/concept/treasure-chest/src/TreasureChest.elm
```diff
@@ -0,0 +1,28 @@
...
+      - a String (for a password)
+
+      - a generic (for the treasure)
+
+-}
+
+
+{-|
+
+1.  define getTreasure
```
jiegillet, line 15: This should be 2.
But really, having this comment and a large `Debug.todo` looks weird, so maybe we could either drop one, or maybe swap

```elm
{-|
2.  define getTreasure
Define getTreasure, with a type annotation, with two parameters:
- a String (for the passwordAttempt)
- a TreasureChest custom type
 and returns:
- if the passwordAttempt does not match the String in the TreasureChest
  - Nothing
- if the passwordAttempt matches the String in the TreasureChest
  - Just treasure (where treasure is the generic in the TreasureChest)
-}

getTreasure passwordAttempt =
  Debug.todo "Please implement the getTreasure function"

```

concepts/generics/links.json
ceddlyburge: I don't think the links in the hints would make sense here. Can you give them another look?

concepts/generics/about.md
ceddlyburge: Hmmm, List.map is a reasonably complex thing to understand, and I think the complexity gets in the way of understanding the idea. And higher order functions are explicitly out of scope in the issue.
Maybe we could add Maybe.map instead?

ceddlyburge:
> I don't really mind calling it map, but I think we should not have a functor concept in the track. For two reasons:

I am uneasy with making students create a Functor, but not telling them what it is. I agree that the term isn't used much in Elm though, and that nearly all of the time you are using Functors in Elm, rather than creating them yourself.

> It's not entirely clear what is a functor and what isn't in Elm. 

There is a formal definition of a Functor, so I think this is clear. Basically it has to have a map function and obey the two functor laws (identity and composition).

How about we compromise, and ask them to do this, and mention that it is a functor (but not go in to depth), and supply a link to some extra reading if they are interested ([Prof Frisbys mostly adequate guide](https://github.com/MostlyAdequate/mostly-adequate-guide) is my favourite).

exercises/concept/treasure-chest/src/TreasureChest.elm
ceddlyburge: > This should be 2.

Interestingly this auto corrects to 1, so I've used words instead, and changed the comments as you suggest.

concepts/generics/about.md
jiegillet: Saying this is confusing because Result exists is the same as saying Pair is confusing because Tuples or records exist.
This code demonstrate that multiple generic types can belong in the same product type or in different variants of a sum type.
But if you prefer not to use it, no worries ^^

concepts/generics/about.md
jiegillet: Sure, Maybe.map is fine too.

concepts/generics/links.json
jiegillet: Specifically I was thinking of https://guide.elm-lang.org/types/custom_types.html, I think it fits quite well. You are right the other don't quite fit.

config.json
```diff
@@ -749,8 +768,12 @@
         "slug": "list-ops",
         "name": "List Ops",
         "uuid": "114b5717-0d82-459e-90c4-f00d8b6beb5b",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "generics"
+        ],
+        "prerequisites": [
+          "generics"
```
jiegillet, line 775: Let's add "lists"

config.json
```diff
@@ -957,8 +980,12 @@
         "slug": "sublist",
         "name": "Sublist",
         "uuid": "8e1cb827-2fa9-46a4-ac50-7f8b3ba66809",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "generics"
+        ],
+        "prerequisites": [
+          "generics"
```
jiegillet, line 987: Let's add "lists" too

exercises/concept/treasure-chest/src/TreasureChest.elm
```diff
@@ -0,0 +1,28 @@
+module TreasureChest exposing (..)
+
+{-| One. define a single variant generic TreasureChest custom type with associated data:
+
+  - a String (for a password)
+
+        - a generic (for the treasure)
```
jiegillet, line 7: The formatting is strange :D

exercises/concept/treasure-chest/src/TreasureChest.elm
```diff
@@ -0,0 +1,28 @@
...
+
+-}
+
+
+type TreasureChest
+    = TreasureChest
```
jiegillet, line 13:
```suggestion
    = TODO
```

jiegillet:
> How about we compromise, and ask them to do this, and mention that it is a functor (but not go in to depth), and supply a link to some extra reading if they are interested ([Prof Frisbys mostly adequate guide](https://github.com/MostlyAdequate/mostly-adequate-guide) is my favourite).

I don't know, I still feel pretty strongly that Evan has always stayed away from using category theory terms for a good reason, category theory is not especially helpful when you are learning Elm, and worse, math can actually scare people away. 

However, you are writing this concept, so I'm happy to leave the design decisions up to you.

exercises/concept/treasure-chest/src/TreasureChest.elm
ceddlyburge: Yep, this is a good idea. This still compiles so elm-format is happy, but the message to the student is clear ...

concepts/generics/about.md
ceddlyburge: I do think Pair is a bad example too actually. I took it from Matthieu's branch where he had started work on this. I think students should get used to reading and writing good and idiomatic Elm code. Familiarity helps with learning. So I'll change the Pair example to something else.

> This code demonstrate that multiple generic types can belong in the same product type or in different variants of a sum type.

I'd like to use a different example than Either, but I don't understand this statement, so I can't propose something different. I don't see a product type in the code, just a sum type / custom type?

I could definitely come up with an example that uses 2

ceddlyburge:
> I don't know, I still feel pretty strongly that Evan has always stayed away from using category theory terms for a good reason, category theory is not especially helpful when you are learning Elm, and worse, math can actually scare people away. However, you are writing this concept, so I'm happy to leave the design decisions up to you.

It would be good if we can come up with something we both like :) And hopefully possible.

I think your position is that you don't want to mention Functors, even in passing.
My position is that I don't want to make a student create a Functor, but not tell them what it is.

And your original point was to create a higher order function.

So maybe we need a higher order function / example that isn't a Functor?

concepts/generics/about.md
ceddlyburge: Being as the other example uses List, I've used List.map after all. I might do another pass later and change everything to Maybe though, to make the examples simpler, so that the Generic stuff becomes more prevalent, and the incidental complexity becomes less relevant.

exercises/concept/treasure-chest/src/TreasureChest.elm
jiegillet: Remove the "-" and rewrite as a normal sentence?
Anything is better than this, it just looks off.

concepts/generics/about.md
jiegillet: What is not idiomatic about `Pair`? It's a record with two generic fields, I think it's a decent example. Of course coming up with something better is fine.

Pair is the example for product type, and Either for the sum type, together they are the basis for all you can build.

ceddlyburge:
How about this? Uses a higher order function, but does not create a functor ...

```elm
multiplyTreasure : (a -> List a) -> TreasureChest a -> TreasureChest (List a)
multiplyTreasure multiplier (TreasureChest password treasure) =
    TreasureChest
        password
        (multiplier treasure)
```

concepts/generics/about.md
ceddlyburge: I have now changed all the examples to use Maybe. I think the signal to noise ratio is better (In this case the signal is the generics, and the noise being whatever the example code is doing.)

concepts/generics/about.md
ceddlyburge: I've changed Pair, as it isn't idiomatic, and it was interesting, as it turns out the record type alias' rarely use generics, so I've added a comment to state that.

> Pair is the example for product type, and Either for the sum type, together they are the basis for all you can build. Of course again, coming up with something better is fine :)

So we have examples of product types and sum types in about.md now, so I think we are good?

> This code demonstrate that multiple generic types can belong in the same product type or in different variants of a sum type.

And I think we are all good with this now as well, I've just used the Result type instead of Either.

So I think there is nothing more to do here, but let me know if I've misunderstood something, I am finding it hard to follow this thread.

ceddlyburge:
Hi @jie, I think we are getting close to agreeing on everything! Hopefully you like the `multiplyTreasure` function, and then we're done :)

concepts/generics/about.md
mpizenberg: > it turns out the record type alias' rarely use generics, so I've added a comment to state that.

Just a remark, that generics in product types are used in every single elm program. And you could easily define a type alias for that record.

```
Browser.sandbox :
    { init : model
    , view : model -> Html msg
    , update : msg -> model -> model
    }
    -> Program () model msg
```

I also personally use them a lot to make parametric views, especially to pass message builders as arguments like

```
someButton : Config msg -> Html msg
someButton msgConfig = generateButtonWith msgConfig
```

This pattern for views is sometimes called the config pattern, as in Evan's sortable table example https://package.elm-lang.org/packages/billstclair/elm-sortable-table/latest/Table#view

That being said, I don't mind at all if you prefer going another route.

ceddlyburge: I remember reading about the config pattern and looking at the code for the sortable table. I remember it being hard to work out what was going on, but interesting. I can't remember the details now though. I think if we want to teach that one of us needs to write a blog post or something, to make it a bit more approachable. I found it tricky to work out what the pattern was, and I've read and written a lot of elm, so I think it would be too much for a beginner.

It's true that the main program for all elm apps is generic, but people very rarely create a type alias for it, and you rarely have to interact with it. If you look at the code for some elm apps on git hub, or at the source code for common elm packages, it is very easy to find a lot of examples of generic type parameters for custom types, and really hard to find examples of generic type parameters for product types / records, especially ones that you could use as an understandable example.

I'm happy to take this statement out, but I think it is true, having looked at a lot of code trying to find a good example!

mpizenberg: Having `Maybe` and `Result` in the core elm package definitely tips the scales to more sum types with generics. However, understanding that generics can be used with both building blocks of elm types (sum and product) is important as a learning lesson. And being able to understand these when reading the types in the package doc or in code is almost as important as actually using them ourselves. So I would not devalue usage of generics for product types in the explanations.

As for the config pattern, it’s actually very simple. It’s just fancy name to say "pass behavior as input argument". In OOP, that’s sometimes called "dependency injection" which is another fancy name. It provides great flexibility when you don’t want to encode in your own function all the possible ways it could be used. You just state that some of the behavior/type has to be provided in the function arguments, in a way that you can use inside your function.

ceddlyburge: Hi Matthieu. about.md does have an example of using generics with records / product types, I've just changed it from Pair (which is code that you would never see in real life), to Event (which is code you might write in real life).

I think it's important to show real practical code examples. If a student sees the Pair code, they might think that it is useful for some problem they are solving and use it, when probably they should be using a Tuple or something instead. Seeing a lot of idiomatic code also means that you get a feel for what id idiomatic in Elm.

Anyway, thanks for the simple explanation for the config pattern.

And is there a change you want making here? If so let's discuss ...

concepts/generics/about.md
```diff
@@ -0,0 +1,93 @@
...
+Being independent does not prevent them from becoming the same type, it just enables them to be different.
+
+```elm
+-- Result is generic over the types error and value
+type Result error value
+    = Ok value
+    | Err error
+```
+
+Record type alias' can also have type parameters, although it is less common.
```
jiegillet, line 42:
```suggestion
Record type aliases can also have type parameters, although it is less common.
```

concepts/generics/about.md
```diff
@@ -0,0 +1,93 @@
...
+-- Event is generic over the type a
+type alias Event a =
+    { at : Time
+    , data : a 
+    }
+```
+
+One important remark about names of type parameters, is that they can be multiple characters long.
+The only constraint is that they must be lowercase names.
+As such, when useful, we tend to give them descriptive names.
```
jiegillet, line 54: I would move the 3 lines above line 31 "A type can be generic over multiple other types..." because we use a longer name for Result, it's weird to mention this after we used it already.

concepts/generics/about.md
```diff
@@ -0,0 +1,93 @@
...
+```elm
+withDefault : a -> Maybe a -> a
+withDefault default maybe =
+    case maybe of
+      Just value -> value
+      Nothing -> default
+```
+
+Functions can take other functions as parameters, and these can also be generic
+Functions that take other functions as parameters are called higher order functions.
```
jiegillet, line 77:
```suggestion
Functions that take other functions as parameters are called higher order functions, and these can also be generic.
```

jiegillet:
Oh sorry, I thought you had introduced it already, so I re-reviewed the whole thing.
Yes, I'm good with `multiplyTreasure` and I'm also good with `about.md` as it is now (after minor tweaks).
Thanks for keeping the momentum in this long PR :p

ceddlyburge:
HI @jie, I think this is finished now, can you check / approve?
Thanks, Cedd

jiegillet (approved):
Looks good 🎉 !
I found two more small details to fix, other than that, it turned out great.

PR #533, opened 2022-11-20
Add new practice exercise `zebra-puzzle`

jiegillet:
I felt like adding this one. It's a hard problem and I was hoping the use of types might help building a clear solution, but in the end, not that much...

In my example implementation I used generics (for example in `sameHouseConstraint`), so we could say it practices generics, but I'm not really sure it's the only approach.

### DIFF
--- a/exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
+++ b/exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
@@ -1,0 +1,321 @@
+module ZebraPuzzle exposing (Nationality(..), drinksWater, ownsZebra)
+
+
+type Number
+    = One
+    | Two
+    | Three
+    | Four
+    | Five
+
+
+type Nationality
+    = Japanese
+    | Norwegian
+    | English
+    | Spanish
+    | Ukrainian
+
+
+type Color
+    = Red
+    | Green
+    | Blue
+    | Yellow
+    | Ivory
+
+
+type Drink
+    = Tea
+    | Milk
+    | OrangeJuice
+    | Coffee
+    | Water
+
+
+type Animal
+    = Zebra
+    | Dog
+    | Snails
+    | Horse
+    | Fox
+
+
+type CigaretteBrand
+    = Chesterfields
+    | OldGold
+    | Kools
+    | LuckyStrike
+    | Parliaments
+
+
+type alias House =
+    { number : Number
+    , nationality : Nationality
+    , color : Color
+    , drink : Drink
+    , animal : Animal
+    , cigaretteBrand : CigaretteBrand
+    }
+
+
+drinksWater : Maybe Nationality
+drinksWater =
+    solution
+        |> List.filter (\house -> .drink house == Water)
+        |> List.map .nationality
+        |> List.head
+
+
+ownsZebra : Maybe Nationality
+ownsZebra =
+    solution
+        |> List.filter (\house -> .animal house == Zebra)
+        |> List.map .nationality
+        |> List.head
+
+
+solution : List House
+solution =
+    let
+        reducedPossibilities =
+            allPossibilities |> applySameHouseConstraints sameHouseConstraints |> applyneighboringHouseConstraint
+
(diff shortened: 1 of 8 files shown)


exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
```diff
@@ -0,0 +1,321 @@
...
+            List.all (\h -> field h /= field house || .number h == .number house) houses
+    in
+    if houseOnlyHasOneTypeOfField || typeOfFieldIsOnlyFoundInHouse then
+        Just (sameHouseConstraint field (field house) .number (.number house))
+
+    else
+        Nothing
+
+
+unique : List a -> List a
```
ceddlyburge, line 314: I think using a Set here is more usual in Elm

exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
```diff
@@ -0,0 +1,321 @@
...
+                possibilities
+
+            else
+                iterate newPossibilities
+    in
+    iterate reducedPossibilities
+
+
+allPossibilities : List House
+allPossibilities =
```
ceddlyburge, line 99: Is there something in List.Extra that makes this easier?

exercises/practice/zebra-puzzle/tests/Tests.elm
```diff
@@ -0,0 +1,20 @@
...
+
+import Expect
+import Test exposing (Test, describe, skip, test)
+import ZebraPuzzle
+
+
+tests : Test
+tests =
+    describe "ZebraPuzzle"
+        [ test "resident who drinks water" <|
```
ceddlyburge, line 11: This code / tests are a bit strange, as the answer is already known, so it would be very easy to pass the tests just by returning these values, which is what you would do if you were using tdd. This is a standard exercise though, right? So probably not something we should worry about here.

ceddlyburge (approved):

exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
jiegillet: Of course, but `Set` won't take a custom types, and I didn't feel like defining some `toComparable` and `fromComparable` for all 5 types of custom types I use this function on. I really wish `Set` and `Dict` could handle any type out of the box.

exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
jiegillet: Mmh, yeah maybe `cartesianProduct` + `transpose` + `List.map5` could make it cleaner. `unique` is also in there.
But we don't have access to `List.Extra` with the standard `elm.json`.

exercises/practice/zebra-puzzle/tests/Tests.elm
jiegillet: Right, this is a standard exercise, and incredibly easy to pass if you cheat.
I was thinking of adding an analyzer check to make sure that the functions are not exactly `drinksWater = Just Norwegian`, but beyond that, I'm not too worried about it.

exercises/practice/zebra-puzzle/.meta/src/ZebraPuzzle.example.elm
ceddlyburge: If its a lot easier, then its probably worth a comment at least, so that a student will know there is a better way, even if we don't do it in the exemplar.

jiegillet: I don't think it's that much easier, even with the functions I suggested you would need a messy pattern match for a list of 5 elements... And using another package will mean the test-runner and the analyzer will not run, so I think it's a suggestion with potential negative consequences. 
 
The way I did it looks messy but it's not complex, it's the same thing repeated 5 times. It's probably fine.

PR #544, opened 2022-12-19
Fix links (the reference style markdown format was wrong)

ceddlyburge:
I've just done the Opaque Types concept exercise on the website, and the markdown links are wrong, so this fixes them up.

@jiegillet @mpizenberg


mpizenberg (approved):
Thanks for the fixes

PR #545, opened 2022-12-19
Add missing reference style link

ceddlyburge:
Just did another check and realised I missed this one.

Something I will look out for on future pull requests! I kind of assumed that the linter would pick this up, but it does not!

@mpizenberg


mpizenberg (approved):
woops ^^

PR #547, opened 2022-12-23
Fix operator descriptions in partial application / composition concept

nidico:


ceddlyburge:
Thanks Nicolas, good spot!

Those pipe operators always confuse me, so I looked it up in the docs and you definitely have it right :)

ceddlyburge (approved):

PR #549, opened 2023-01-08
Add new practice exercise `dnd-character`

jiegillet:
I'm adding this one in preparation/research for the `random` concept.

I would love to have your opinion on the testing. The first group of tests are for a pure function, but the last two test the output of `Generator`s.
@ceddlyburge I know you are on holiday. This is not urgent :)

@mpizenberg we should still avoid fuzzers, right? What I did instead was to generate 1000 values from a specific seed and tested that some properties held for all. The library warns against using `fromGenerator` anyway, so maybe it's not too bad.

The first generator `ability` is suppose to give the sum of the largest 3 dice out of 4. The canonical data indicated to test whether all values the correct range (>= 3 and <= 18), but I added one about the distribution (3 is a less likely outcome than 4, 4 less likely  than 5, 4 *way* less likely than 12, 17 more likely than 18...).

The second generator `character` gives out a character with 6 abilities + `hitpoints` directly related to one of the abilities. The canonical data indicated to test that all characters are valid (abilities in the correct range + correct `hitpoints`) and also check that the individual abilities are only calculated once. That last one makes no sense for generators, so I dropped it, but I added more tests for checking that not all characters are identical and that not all abilities are identical.

Does that sounds like a valid way to test generators? If so, I might add a couple more practice exercises before adding the concept.

### DIFF
--- a/config.json
+++ b/config.json
@@ -10,7 +10,7 @@
     "highlightjs_language": "elm"
   },
   "test_runner": {
-    "average_run_time": 2
+    "average_run_time": 2.0
   },
   "status": {
     "concept_exercises": true,
--- a/config.json
+++ b/config.json
@@ -10,7 +10,7 @@
     "highlightjs_language": "elm"
   },
   "test_runner": {
-    "average_run_time": 2.0
+    "average_run_time": 2
   },
   "status": {
     "concept_exercises": true,
@@ -1130,6 +1130,17 @@
           "records"
         ],
         "difficulty": 8
+      },
+      {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "c22b805f-36db-407c-92d1-39ba0e4fa041",
+        "practices": [],
+        "prerequisites": [
+          "lists",
+          "records"
+        ],
+        "difficulty": 5
       }
     ]
   }
--- a/exercises/practice/dnd-character/.docs/instructions.md
+++ b/exercises/practice/dnd-character/.docs/instructions.md
@@ -1,0 +1,31 @@
+# Instructions
+
+For a game of [Dungeons & Dragons][dnd], each player starts by generating a character they can play with.
+This character has, among other things, six abilities; strength, dexterity, constitution, intelligence, wisdom and charisma.
+These six abilities have scores that are determined randomly.
+You do this by rolling four 6-sided dice and record the sum of the largest three dice.
+You do this six times, once for each ability.
+
+Your character's initial hitpoints are 10 + your character's constitution modifier.
+You find your character's constitution modifier by subtracting 10 from your character's constitution, divide by 2 and round down.
+
+Write a random character generator that follows the rules above.
+
+For example, the six throws of four dice may look like:
+
+- 5, 3, 1, 6: You discard the 1 and sum 5 + 3 + 6 = 14, which you assign to strength.
+- 3, 2, 5, 3: You discard the 2 and sum 3 + 5 + 3 = 11, which you assign to dexterity.
+- 1, 1, 1, 1: You discard the 1 and sum 1 + 1 + 1 = 3, which you assign to constitution.
+- 2, 1, 6, 6: You discard the 1 and sum 2 + 6 + 6 = 14, which you assign to intelligence.
+- 3, 5, 3, 4: You discard the 3 and sum 5 + 3 + 4 = 12, which you assign to wisdom.
+- 6, 6, 6, 6: You discard the 6 and sum 6 + 6 + 6 = 18, which you assign to charisma.
+
+Because constitution is 3, the constitution modifier is -4 and the hitpoints are 6.
+
+## Notes
+
+Most programming languages feature (pseudo-)random generators, but few programming languages are designed to roll dice.
+One such language is [Troll][troll].
+
+[dnd]: https://en.wikipedia.org/wiki/Dungeons_%26_Dragons
+[troll]: https://di.ku.dk/Ansatte/?pure=da%2Fpublications%2Ftroll-a-language-for-specifying-dicerolls(84a45ff0-068b-11df-825d-000ea68e967b)%2Fexport.html
--- a/exercises/practice/dnd-character/.meta/src/DndCharacter.example.elm
+++ b/exercises/practice/dnd-character/.meta/src/DndCharacter.example.elm
@@ -1,0 +1,56 @@
+module DndCharacter exposing (Character, ability, character, modifier)
+
+import Random exposing (Generator)
+
+
+type alias Character =
+    { strength : Int
+    , dexterity : Int
+    , constitution : Int
+    , intelligence : Int
+    , wisdom : Int
+    , charisma : Int
+    , hitpoints : Int
+    }
+
+
+modifier : Int -> Int
+modifier score =
+    floor ((toFloat score - 10) / 2)
+
+
+ability : Generator Int
+ability =
+    let
+        sumOfMaxThree =
+            List.sort >> List.reverse >> List.take 3 >> List.sum
+    in
+    Random.list 4 (Random.int 1 6)
+        |> Random.map sumOfMaxThree
+
+
+character : Generator Character
+character =
+    Random.constant Character
+        |> apply ability
+        |> apply ability
+        |> apply ability
+        |> apply ability
+        |> apply ability
+        |> apply ability
+        |> Random.andThen
+            (\hitpointsToCharacter ->
+                let
+                    dummyChar =
+                        hitpointsToCharacter 0
+
+                    hitpoints =
+                        10 + modifier dummyChar.constitution
+                in
+                Random.constant (hitpointsToCharacter hitpoints)
+            )
+
+
+apply : Generator a -> Generator (a -> b) -> Generator b
+apply genA =
+    Random.andThen (\aToB -> Random.map aToB genA)
--- a/exercises/practice/dnd-character/tests/Tests.elm
+++ b/exercises/practice/dnd-character/tests/Tests.elm
@@ -1,0 +1,134 @@
+module Tests exposing (tests)
+
+import DndCharacter exposing (Character)
+import Expect
+import Random exposing (Generator, Seed)
+import Test exposing (Test, describe, skip, test)
+
+
+generateValuesFromSeed : Int -> Generator a -> Seed -> List a
+generateValuesFromSeed size generator seed =
+    if size <= 0 then
+        []
+
+    else
+        let
+            ( value, nextSeed ) =
+                Random.step generator seed
+        in
+        value :: generateValuesFromSeed (size - 1) generator nextSeed
+
+
+tests : Test
+tests =
+    describe "DndCharacter"
+        [ describe "ability modifier"
+            [ test "ability modifier for score 3 is -4" <|
+                \() ->
+                    DndCharacter.modifier 3
+                        |> Expect.equal -4
+            , skip <|
+                test "ability modifier for score 4 is -3" <|
+                    \() ->
+                        DndCharacter.modifier 4
+                            |> Expect.equal -3
+            , skip <|
+                test "ability modifier for score 5 is -3" <|
+                    \() ->
+                        DndCharacter.modifier 5
+                            |> Expect.equal -3
+            , skip <|
+                test "ability modifier for score 6 is -2" <|
+                    \() ->
+                        DndCharacter.modifier 6
+                            |> Expect.equal -2
+            , skip <|
+                test "ability modifier for score 7 is -2" <|
+                    \() ->
+                        DndCharacter.modifier 7
+                            |> Expect.equal -2
+            , skip <|
+                test "ability modifier for score 8 is -1" <|
+                    \() ->
+                        DndCharacter.modifier 8
+                            |> Expect.equal -1
+            , skip <|
+                test "ability modifier for score 9 is -1" <|
+                    \() ->
+                        DndCharacter.modifier 9
+                            |> Expect.equal -1
+            , skip <|
+                test "ability modifier for score 10 is 0" <|
+                    \() ->
+                        DndCharacter.modifier 10
+                            |> Expect.equal 0
+            , skip <|
+                test "ability modifier for score 11 is 0" <|
+                    \() ->
+                        DndCharacter.modifier 11
+                            |> Expect.equal 0
+            , skip <|
+                test "ability modifier for score 12 is +1" <|
+                    \() ->
+                        DndCharacter.modifier 12
+                            |> Expect.equal 1
+            , skip <|
+                test "ability modifier for score 13 is +1" <|
+                    \() ->
+                        DndCharacter.modifier 13
+                            |> Expect.equal 1
+            , skip <|
+                test "ability modifier for score 14 is +2" <|
+                    \() ->
+                        DndCharacter.modifier 14
+                            |> Expect.equal 2
+            , skip <|
+                test "ability modifier for score 15 is +2" <|
+                    \() ->
+                        DndCharacter.modifier 15
+                            |> Expect.equal 2
+            , skip <|
+                test "ability modifier for score 16 is +3" <|
+                    \() ->
+                        DndCharacter.modifier 16
+                            |> Expect.equal 3
+            , skip <|
+                test "ability modifier for score 17 is +3" <|
+                    \() ->
+                        DndCharacter.modifier 17
+                            |> Expect.equal 3
+            , skip <|
+                test "ability modifier for score 18 is +4" <|
+                    \() ->
+                        DndCharacter.modifier 18
+                            |> Expect.equal 4
+            ]
+        , skip <|
+            test "random ability is within range" <|
+                \() ->
+                    generateValuesFromSeed 1000 DndCharacter.ability (Random.initialSeed 42)
+                        |> List.all (\score -> 3 <= score && score <= 18)
+                        |> Expect.equal True
+        , skip <|
+            test "random character is valid" <|
+                \() ->
+                    let
+                        validCharacter { strength, dexterity, constitution, intelligence, wisdom, charisma, hitpoints } =
+                            (strength >= 3)
+                                && (strength <= 18)
+                                && (dexterity >= 3)
+                                && (dexterity <= 18)
+                                && (constitution >= 3)
+                                && (constitution <= 18)
+                                && (intelligence >= 3)
+                                && (intelligence <= 18)
+                                && (wisdom >= 3)
+                                && (wisdom <= 18)
+                                && (charisma >= 3)
+                                && (charisma <= 18)
+                                && (hitpoints == 10 + DndCharacter.modifier constitution)
+                    in
+                    generateValuesFromSeed 1000 DndCharacter.character (Random.initialSeed 42)
+                        |> List.all validCharacter
+                        |> Expect.equal True
+        ]
--- a/exercises/practice/dnd-character/tests/Tests.elm
+++ b/exercises/practice/dnd-character/tests/Tests.elm
@@ -1,8 +1,10 @@
 module Tests exposing (tests)
 
+import Dict
 import DndCharacter exposing (Character)
 import Expect
 import Random exposing (Generator, Seed)
+import Set
 import Test exposing (Test, describe, skip, test)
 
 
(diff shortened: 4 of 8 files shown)


config.json
```diff
@@ -1130,6 +1130,17 @@
           "records"
         ],
         "difficulty": 8
+      },
+      {
+        "slug": "dnd-character",
+        "name": "D&D Character",
+        "uuid": "c22b805f-36db-407c-92d1-39ba0e4fa041",
+        "practices": [],
```
jiegillet, line 1138: I will add `"random"` here when it's added.

mpizenberg (commented):
It's an interesting practice exercise.

There are two reasons why we didn't allow fuzz testing in exercism tests:

1. Because their structure were not detected by the test parser creating the reports automatically. Maybe this one is not valid anymore, I kinda remember that you helped Cedd improve with these parsers so maybe its fine now?
2. Because it degrades the error reporting for the student in two ways, one way is we don't know in advance the error that may occur, and the second way is that we have less control in the actual error message.

By generating yourself the 1000 rolls test generators you bypass problem (1), but your tests still kinda have not great error messages. Some can easily be improved. For example, I'd suggest removing the `Expect.all` factor out some logic (like the count function) and have a different test for each expectation, with a good error message if that happen.

In general, try figuring out a way to have as clear as possible error messages.

For some tests, like "random character has independent abilities", I'd love a doc comment explaining where the chosen value (4) comes from. Ideally, with a computation explaining how the different parameters we chose make sure that a correct implementation has 99.99% chance of passing. We don't want frequent issues for students when they have to change their code, just to relaunch the tests. A note on that subject for mentors might come in handy by the way for the very rare encounter where the student code is actually correct. I don't remember if there is such a way to add notes for mentors.

Final note, your distribution testing made me think a lot about the new feature of the new fuzz library: https://package.elm-lang.org/packages/elm-explorations/test/latest/Test-Distribution
I know I said we shouldn't use fuzz tests, but just in case, you might want to try it at home and see if using it would be valuable enough to bend the rule, if possible.

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,167 @@
...
+generateValuesFromSeed size generator seed =
+    if size <= 0 then
+        []
+
+    else
+        let
+            ( value, nextSeed ) =
+                Random.step generator seed
+        in
+        value :: generateValuesFromSeed (size - 1) generator nextSeed
```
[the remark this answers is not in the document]
ceddlyburge: I think (1) is still a problem, but I think it would be fairly easy to sort out, and I would be happy to do it if it makes life better.

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,167 @@
...
+
+import Dict
+import DndCharacter exposing (Character)
+import Expect
+import Random exposing (Generator, Seed)
+import Set
+import Test exposing (Test, describe, skip, test)
+
+
+generateValuesFromSeed : Int -> Generator a -> Seed -> List a
```
ceddlyburge, line 11: I would probably hard code ability and character versions of this, to save computation time and make the tests a tiny bit less complicated.

ceddlyburge (commented):
This looks good to me. I second Matthieu's comments about good test failure messages being important, although I haven't worked out how good / bad they are at the moment. I think Matthieu analysed this though ...

jiegillet:
Thank you both for the reviews. 
I agree that my error messages were particularly bad :D

I spent some time testing out better tests error-wise and also tried out `Fuzz` functions as an experiment. Some random observations.

## Test parser
Currently `fuzz` and `fuzzWith` are not detected, so a change would be required on the test parser. Not a big deal, the hardest question would be what to show the students. 
Currently from `test "name" <| \() -> inner` we show `inner` and `name` separately. 
For `fuzz  fuzzer "name" <| \generated -> inner generated` we may need to show the whole thing and then `name` would be duplicated.

## Fuzzers have changed in elm-explorations/test v2
I believe the most significant changes in the new version of `elm-explorations/test` is around fuzzers, so I'm hesitant to introduce fuzzers before upgrading to the new version.
A big change was the introduction of the distribution checks, those look pretty tasty, but I haven't tried them, and I don't know what the messages would look like.

## Fuzzers from Generators
Usually, fuzzers are used for property testing, although `Fuzz.custom` (v1) / `Fuzz.fromGenerator` (v2) can use `Random.Generator`. Both functions are not encouraged because the shrinking behavior may not work well.
For this particular exercise though, shrinking is not an issue, because for `ability` and `character` returns a value that doesn't need shrinking, unlike a large list of random values or something.

## Default fuzz runs is not enough
For the specific purpose of testing distributions, the default number of `fuzz` generated values is 100, which was not enough in my explorations, so instead I used `fuzzWith { runs = 1000 }`.

## 99.99% chance of passing
> For some tests, like "random character has independent abilities", I'd love a doc comment explaining where the chosen value (4) comes from. Ideally, with a computation explaining how the different parameters we chose make sure that a correct implementation has 99.99% chance of passing. We don't want frequent issues for students when they have to change their code, just to relaunch the tests. 

This is a cool idea. I originally picked the value of 5, then went back to 4 to address this concern out of a gut feeling, but I didn't crunch the numbers. Finding exact values will be a pain, but I can certainly come up with estimations.

> A note on that subject for mentors might come in handy by the way for the very rare encounter where the student code is actually correct. I don't remember if there is such a way to add notes for mentors.

The notes for mentors can be added [here](https://github.com/exercism/website-copy/tree/main/tracks/elm/exercises). I see Cedd has created a few already, nice 👍 

## Comparing  errors
Here are two equivalent tests that I optimized for readability and clearest errors, using generated values vs. fuzzers

```elm
        [ fuzzWith { runs = 1000 } (Fuzz.custom DndCharacter.ability Shrink.noShrink) "fuzz: random ability should be at least 3" <|
                \ability -> Expect.atLeast 3  ability
        , test "minimum ability should be 3" <|
                \() ->
                    generateValuesFromSeed 1000 DndCharacter.ability (Random.initialSeed 42)
                        |> List.minimum
                        |> Expect.equal (Just 3) 
        , ...
```

Errors for a faulty implementations (dice that can return 0 to 6), the errors were
```
Given 2

    2
    ╷
    │ Expect.atLeast
    ╵
    3


Given 0

    0
    ╷
    │ Expect.atLeast
    ╵
    3
```
and
```
    Just 1
    ╷
    │ Expect.equal
    ╵
    Just 3
```

Much better than the `True Expect.equal False` I had before :D
I actually really like the fuzzers errors, because they show different failure cases with no repetition. It would be messy to emulate in the other test. 

## Conclusion
My conclusion is... More data required.
The v2 tests introduction of distributions could be really nice for this exercise but also for the random concept, and I want to try them out before making a decision. 
Since we already have an issue open for switching to v2, I will close that one first, and then come back to this exercise.

ceddlyburge:
Thanks @jiegillet , very thorough and impressive analysis!

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+                fuzzWith
+                    { runs = 10000
+                    , distribution =
+                        Test.expectDistribution
+                            [ ( Test.Distribution.moreThanZero, "is 3", (==) 3 ) ]
+                    }
+                    (Fuzz.fromGenerator DndCharacter.ability)
+                    "generated ability should sometimes be 3"
+                <|
+                    \_ -> Expect.pass
```
jiegillet, line 118: This is a bit more tricky. In this case we are not testing a property of the ability (the test is `Expect.pass`) but instead we are looking at the distribution. We want to make sure that the generator _sometimes_ produces a 3. `( Test.Distribution.moreThanZero, "is 3", (==) 3 )` means that we want one ability to be 3 at least once. If it's never 3, the assumption we had on the distribution is false and the test will fail.

However, getting a 3 is quite rare (0.1%), so we need to take a very large sample to make sure that we get one at least 99.99% of the time. So I crunched the numbers.

Given a specific score that appears with probability $p$, the probability that it doesn't appear is $1 - p$, the probability that it doesn't appear $n$ times in a row is $(1 - p)^n$, so the probability that it does appear at least once in $n$ runs is:
```math
1 - (1 - p)^n
```

Screenshot 2023-10-07 at 21 54 44

For 3, we need 10k runs to be safe, for 18 we need 1k.
In practice, I tested getting a 3 with 1k runs and I never got it to fail, so maybe I'm missing something, possibly related to how `Test.expectDistribution` works, the docs say:
> Uses a statistical test to make sure the distribution doesn't pass or fail the distribution by accident (a flaky test). Will run more tests than specified with the runs config option if needed.

An error for a test of this type looks like
```
    Distribution report:
    ====================
      <uncategorized>:  100%  (10000x)  ██████████████████████████████
      is 3:               0%      (0x)  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
    
    Distribution of label "is 3" was insufficient:
      expected:  more than 0%
      got:       0.000%.
    
    (Generated 10000 values.)
```

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+                    , distribution =
+                        Test.expectDistribution
+                            [ ( Test.Distribution.atLeast 10, "is 13", (==) 13 )
+                            , ( Test.Distribution.atLeast 85, "is not 13", (/=) 13 )
+                            ]
+                    }
+                    (Fuzz.fromGenerator DndCharacter.ability)
+                    "13 has an approximate 13% chance of being picked"
+                <|
+                    \_ -> Expect.pass
```
jiegillet, line 142: This one is also a test on the distribution. 

I wanted to test the actual probability of some abilities to appear. I picked 13 because it was the most common number, but I could add a couple more if you think it's worth doing.
The probability of 13 to appear is about 13.2%, so the probability that something else appears is 86.8%. When testing actual probabilities of events to happen, the docs say:
>Uses a statistical test to make sure the distribution doesn't pass or fail the distribution by accident (a flaky test). Will run more tests than specified with the runs config option if needed.
> 
> This has the consequence of running more tests the closer your expected distribution is to the true distribution. You can thus minimize and speed up this "making sure" process by requesting slightly less % of your distribution than needed.

So I used a window, anything between 10% and 15% (or actually, 85% of not having 13). Now again, for such a window, I needed to make sure to have enough runs so that 99.99% of the time it would work and I calculated the binomial distribution.
In simple terms, the binomial distribution can tell us the probability that an event will happen at least some percentage of the time, given the probability $p$ of the single event to happen and a number of runs $n$. Here are the numbers

Screenshot 2023-10-07 at 22 07 01

So in this case again we need 10k runs.

An error for a test of this type looks like
```
    Distribution report:
    ====================
      is not 13:  90.6%  (127810x)  ███████████████████████████░░░
      is 13:       9.4%   (13260x)  ███░░░░░░░░░░░░░░░░░░░░░░░░░░░
    
    Distribution of label "is 13" was insufficient:
      expected:  10.000%
      got:       9.400%.
    
    (Generated 141070 values.)
```
(that's quite a lot more values than specified, but it ran in 239 ms)

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+            ]
+        , describe "character"
+            [ skip <|
+                fuzzWith { runs = 100, distribution = Test.noDistribution }
+                    (Fuzz.fromGenerator DndCharacter.character)
+                    "generated character strength should be within range"
+                <|
+                    \character ->
+                        Expect.all [ Expect.atLeast 3, Expect.atMost 18 ] character.strength
```
jiegillet, line 151: This is checking that the generated abilities are within range. I did one test per ability for clarity.

An error for a test of this type looks like
```
Given { charisma = 16, constitution = 12, dexterity = 14, hitpoints = 11, intelligence = 12, strength = 21, wisdom = 15 }

    21
    ╷
    │ Expect.atMost
    ╵
    18
```

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+                    "generated character abilities are not all equal"
+                <|
+                    \character ->
+                        let
+                            uniqueAbilities { strength, dexterity, constitution, intelligence, wisdom, charisma } =
+                                [ strength, dexterity, constitution, intelligence, wisdom, charisma ]
+                                    |> Set.fromList
+                                    |> Set.size
+                        in
+                        Expect.atLeast 2 (uniqueAbilities character)
```
jiegillet, line 206: This one is checking that not all abilities are copies of each other. The highest most probable state with all the same states are all 13s ($p=13.2$%), which can happen with probability $p^6 = 10^{-11}$, very unlikely, so 100 runs is fine.
I'm pretty sure I could require having 3 separate values in 100 runs, but the calculations become more complex, so I got lazy and played it safe :)

An error for a test of this type looks like
```
Given { charisma = 10, constitution = 10, dexterity = 10, hitpoints = 10, intelligence = 10, strength = 10, wisdom = 10 }

    1
    ╷
    │ Expect.atLeast
    ╵
    2
```

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+                            [ ( Test.Distribution.atLeast 40, "has low charisma", \char -> char.charisma <= 12 )
+                            , ( Test.Distribution.atLeast 40, "has high charisma", \char -> char.charisma > 12 )
+                            , ( Test.Distribution.atLeast 40, "has low strength", \char -> char.strength <= 12 )
+                            , ( Test.Distribution.atLeast 40, "has high strength", \char -> char.strength > 12 )
+                            ]
+                    }
+                    (Fuzz.fromGenerator DndCharacter.character)
+                    "generated characters are not all equal"
+                <|
+                    \_ -> Expect.pass
```
jiegillet, line 221: Finally, this one check that we don't always generate the same character.

I looked at the cumulative probabilities of each score, and there is a 51.2% chance of getting a 12 or less, so I required the distribution to have at least 40% of 1-12 and at least 40% of 13-18. The binomial distribution gave me a `10^{-12}` probability to have less than 40% in 1000 runs.

An error for a test of this type looks like
```
    Distribution report:
    ====================
      has low charisma:                    100%  (100x)  ██████████████████████████████
      has low strength:                    100%  (100x)  ██████████████████████████████
      has high charisma:                     0%    (0x)  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
      has high strength:                     0%    (0x)  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
    
    Combinations (included in the above base counts):
      has low charisma, has low strength:  100%  (100x)  ██████████████████████████████
    
    Distribution of label "has high charisma" was insufficient:
      expected:  40.000%
      got:       0.000%.
    
    (Generated 100 values.)
```

jiegillet (commented):
I'm back to this PR after almost 10 months of absence!!

I decided to fully embrace the latest version of the tests and try to use the fuzzers. I believe that the error messages can be quite clear if the scope of the test is restrained enough. I gave lots of examples of error messages in comments below.

One thing that we need before merging this is changing the parser, because right now, it can only detect the `describe` and `test`, not the fuzzers, so if this get approved, I will make a PR over at `elm-test-runner`.

I may have gone a little bit overboard with this, I was nerding out and exploring the possibilities of the test distributions, so please let me know if this would be too much for the students. I also hope my math checks out.

The first thing I did was generate (not calculate) the different probabilities of the individual ability scores. The test library made it super easy, all I had to do was add this test
```elm
                fuzzWith
                    { runs = 1000000
                    , distribution =
                        List.range 3 18
                            |> List.map (\n -> ( String.fromInt n, (==) n ))
                            |> Test.reportDistribution
                    }
                    (Fuzz.fromGenerator DndCharacter.ability)
                    "generate all probabilities"
                <|
                    \_ -> Expect.pass
```
And it showed me all the probabilities. I cleaned it up a bit and here they are

Screenshot 2023-10-07 at 22 54 55

ceddlyburge:
Hi Joe, looks like you've really gone big!

I'm in holiday at the moment, I'll give it a look when I get back.

Cheers, Cedd

exercises/practice/dnd-character/tests/Tests.elm
```diff
@@ -0,0 +1,223 @@
...
+                        DndCharacter.modifier 18
+                            |> Expect.equal 4
+            ]
+        , describe "ability"
+            [ skip <|
+                fuzzWith { runs = 1000, distribution = Test.noDistribution }
+                    (Fuzz.fromGenerator DndCharacter.ability)
+                    "generated ability should be at least 3"
+                <|
+                    \ability -> Expect.atLeast 3 ability
```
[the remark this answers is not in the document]
jiegillet: Actually I think it's a pretty good error message, with the test name as context "generated ability should be at least 3"

exercises/practice/dnd-character/tests/Tests.elm
jiegillet: > expectDistribution is for validating that the fuzzer passes the values you want in to the function to be tested, right?

Yes, that's exactly it.

> So I'm not sure this test is doing the right thing. Can you get it to fail by making `DndCharacter.ability` invalid.

Yes, all the example errors I showed are real, and obtained by tweaking the generators. In this case, all you need to change is to make the dice roll from 2 to 6 instead of from 1 to 6 and 3 can never be obtained.

> Maybe it is possible to use Test.Distribution to enforce the score of 3 as well, although I haven't tried it, and it looks tricky, as its a bit magic, and you need to have 4 random inputs in a row all be small. It would be good if it was possible though, and then we could use fail or onFail again to provide a good message.

I think what you are saying is that you would like a way to discard all the generated values that you are not interested in, and then run only the ones you want through the expectation. Is that right? As far as I know, it's not possible.
Or are you saying that you want to manipulate the generator to get particular values? That definitely sounds impossible.

In any case, I think the error message isn't too bad here. The test name is "generated ability should sometimes be 3" and the error message says " Distribution of label "is 3" was insufficient:".

> If we can get these tests perfect, it would also be a good learning resource for people to reference / copy paste when they do their own Random things, so we could also mention that in the instructions (if we can get them to a place where this makes sense).

I don't think a practice exercise is the right place for that. I have been thinking about a `test` concept where we wold explain how tests are structured in Elm. But testing that would be hard :)

exercises/practice/dnd-character/.meta/src/DndCharacter.example.elm
```diff
@@ -0,0 +1,58 @@
...
+        |> apply ability
+        |> Random.andThen addHitpoints
+
+
+apply : Generator a -> Generator (a -> b) -> Generator b
+apply genA =
+    Random.andThen (\aToB -> Random.map aToB genA)
+
+
+addHitpoints : (Int -> Character) -> Generator Character
```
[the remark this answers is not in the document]
jiegillet: I simplified it greatly in [bcf327d](https://github.com/exercism/elm/pull/549/commits/bcf327de1ba4d458247248e0d4f3d4956a861b0f)

exercises/practice/dnd-character/.docs/instructions.md
```diff
@@ -0,0 +1,31 @@
+# Instructions
```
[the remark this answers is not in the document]
jiegillet: We can't modify the instructions, but do you think it's unclear, given the stub file?
`modifier` is the one that's the least clear, but it has so many tests...

exercises/practice/dnd-character/tests/Tests.elm
ceddlyburge: There is definitely something I don't understand about this, but maybe its best not to worry. It sounds like it is doing the right thing.

Maybe the test messages are ok too. They are complex, but that is something a student would have to get used to when working with random, so probably good to expose it.

exercises/practice/dnd-character/tests/Tests.elm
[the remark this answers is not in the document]
ceddlyburge: The `Given 2` bit of the message seems a bit strange, but yes, you're right, it probably makes sense to use the standard elm test message output.

exercises/practice/dnd-character/.docs/instructions.md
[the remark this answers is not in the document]
ceddlyburge: If its part of the standard instructions, lets just leave it ...

ceddlyburge (approved):
This looks good, I think my only remaining comment is that the `Given` part of the test messages look strange. It isn't the input to the system under test, as you would expect with a given / when. then, its more the output from the system under test.

jiegillet:
> This looks good, I think my only remaining comment is that the `Given` part of the test messages look strange. It isn't the input to the system under test, as you would expect with a given / when. then, its more the output from the system under test.

Well, it's both output and input. It's the output of the fuzzer used to generate data, and the input to the actual test. I found that it's the most clear in 
```elm
                fuzzWith { runs = 100, distribution = Test.noDistribution }
                    (Fuzz.fromGenerator DndCharacter.character)
                    "generated character strength should be within range"
                <|
                    \character ->
                        Expect.all [ Expect.atLeast 3, Expect.atMost 18 ] character.strength
```
with the error
```
Given { charisma = 10, constitution = 10, dexterity = 10, hitpoints = 10, intelligence = 10, strength = 10, wisdom = 10 }

    1
    ╷
    │ Expect.atLeast
    ╵
    2
```
since here the input is a character and the expectation a number.

exercises/practice/dnd-character/tests/Tests.elm
ceddlyburge: It might be good to have a testing concept, where we ask the student to write some tests, and then this could rely on that concept, and then students should be able to read the messages.
Perhaps we could update the test runner to also extract and run tests from the students code to make it happen?

ceddlyburge:
> Well, it's both output and input. It's the output of the fuzzer used to generate data, and the input to the actual test. I found that it's the most clear in

Its the input to the test, but not the input to the function being tested (DndCharacter.character), which is what it normally would be, and what I would generally want to see. I think this is the main point of confusion for me. Ideally what you want to see is the "Random" inputs, so you could infer why your function wasn't doing what it was supposed to. Not sure if this is a possible thing though ...

jiegillet:
I think I see what you mean, you want to see the random integer inside the `Seed` and see how it propagates and generates a value through the generator. But that's impossible because generators are opaque types, they are not normal functions. It would also be almost impossible to follow anyway, I don't think it would help.

But that's the deal with generators and JSON decoders, they are a little bit of a magic black box.

mpizenberg:
I've seen many commits related to fuzzed tests but honestly I haven't taken the time to look at it. If you can TLDR me on Elm slack DMs that would be ideal.

ceddlyburge:
> I think I see what you mean, you want to see the random integer inside the Seed and see how it propagates and generates a value through the generator. But that's impossible because generators are opaque types, they are not normal functions. It would also be almost impossible to follow anyway, I don't think it would help.

Yeah, ideally you want to see the random numbers that get fed in to the function, so then you can reason about what happened.

I've never used Random or tried to test it, so not sure what the conventions / idiomatic way of doing things is, but its interesting stuff, I might look in to it, or put something on the Elm slack or something like that

jiegillet:
I actually wrote a duplicate of Elm's random module for Gleam (although I never finished it not published it) so I'm pretty familiar with the internals, and while it's not extremely complex, it's definitely too complex to help with debugging and reasoning about.
These types of monads/generators/decoders are very powerful, but they need some experience to get a feel for it and be more effective at debugging.

PR #551, opened 2023-01-15
Update `elm-explorations/test`

jiegillet:
Part of https://github.com/exercism/elm-test-runner/issues/42.

I had to make changes to the tests of 4 exercises: `concept/ticket-please/tests`, `concept/treasure-factory`, `practice/custom-set/tests` and `practice/rna-transcription`. The reason is that `Expect.true "reason"` disappeared, and had to be replaced with `Expect.equal True |> Expect.onFail "reason"`.

The test runner download and uses the template file of this repo (on the master branch) when rebuilding, so what needs to be done now is:
1. Prepare a PR in elm-test-runner upgrading elm-test-rs and the smoke test examples
3. Test locally that the test runner will be fine with this branch
4. Take a leap of faith and merge this PR
5. Merge the PR from elm-test-runner, that should trigger a re-build of the docker container with the correct template

I'm a bit worried of things going wrong, especially since we cannot merge things directly in elm-test-runner. 

Maybe building elm-test-runner with a local copy of the elm.json instad of downloading this one could help?

### DIFF
--- a/elm-tooling.json
+++ b/elm-tooling.json
@@ -2,6 +2,6 @@
     "tools": {
         "elm": "0.19.1",
         "elm-format": "0.8.5",
-        "elm-test-rs": "2.0.0"
+        "elm-test-rs": "3.0.0"
     }
 }
--- a/exercises/concept/annalyns-infiltration/elm.json
+++ b/exercises/concept/annalyns-infiltration/elm.json
@@ -17,12 +17,13 @@
     },
     "test-dependencies": {
         "direct": {
-            "elm-explorations/test": "1.2.2",
-            "rtfeldman/elm-iso8601-date-strings": "1.1.3"
+            "elm-explorations/test": "2.1.0",
+            "rtfeldman/elm-iso8601-date-strings": "1.1.4"
         },
         "indirect": {
+            "elm/bytes": "1.0.8",
             "elm/html": "1.0.0",
-            "elm/virtual-dom": "1.0.2"
+            "elm/virtual-dom": "1.0.3"
         }
     }
 }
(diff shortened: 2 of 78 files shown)


jiegillet:
Ha, unexpectedly the CI with the current test-runner actually passes all exercises, I didn't realize that `Expect.onFail` was already there in v1.

That's perfect, it makes this PR mergeable even with the old test runner.
Then all we need to do is make sure that the PR over there is fine. Of course we should still merge those close to each other so that nothing else triggers a rebuild of the docker image (like a dependabot update).

ceddlyburge (approved):
This looks good to me, and actually the rna-transcription code has improved a bit.

The changes make me wonder if we should do something about all the elm.json files sometime? Maybe have a template one and copy it around as part of the build or something like that?

What do you think? It doesn't need doing now but I could create an issue for it ...

jiegillet:
> The changes make me wonder if we should do something about all the elm.json files sometime? Maybe have a template one and copy it around as part of the build or something like that?
> 
> What do you think? It doesn't need doing now but I could create an issue for it ...

I'm not too sure it would help that much, no need for an issue now. We are always going to have co-dependencies between the test runner and the code, it would only be a trade-off for some other unforeseen trouble :)

mpizenberg (approved):
All look good to me too

mpizenberg:
PS, someone just asked about this thing on the exercism forum https://forum.exercism.org/t/running-elm-test-after-following-elm-track-docs-has-a-version-deiscrepancy/2917

I received an email notif today because I config my email subscriptions on the forum for the Elm category. If you guys could also subscribe, that would be awesome.

jiegillet:
Good idea, done!

ceddlyburge:
Think I've done that as well.

PR #552, opened 2023-02-01
Update about.md

AJEdgarCraft:
The Dict.map function does not return Dict.empty, change the function name from "empty".

The Dict.map annotation is (k -> a -> b) -> Dict k a -> Dict k b, but Dict k a (alice) is missing from the function.


ceddlyburge (approved):
Hi There, good spot! And thanks for the correction.

AJEdgarCraft:
Thank you for supporting beginners. 🍻

PR #565, opened 2023-04-07
corrected wrong test case

debanwita27:
"Casting a spell spends double the mana" 
mana = 20 and manaCost = 9
updated mana should be 20 - 2*9 = 20 - 18 = 2
test case given : 11(wrong)

### DIFF
--- a/exercises/concept/role-playing-game/tests/Tests.elm
+++ b/exercises/concept/role-playing-game/tests/Tests.elm
@@ -33,7 +33,7 @@ tests =
             [ test "Casting a spell causes damage of double the mana" <|
                 \() ->
                     Expect.equal (castSpell 9 { name = Nothing, level = 10, health = 69, mana = Just 20 })
-                        ( { name = Nothing, level = 10, health = 69, mana = Just 2 }, 18 )
+                        ( { name = Nothing, level = 10, health = 69, mana = Just 11 }, 18 )
             , test "Attempting to cast a spell with insufficient mana does nothing" <|
                 \() ->
                     Expect.equal (castSpell 39 { name = Nothing, level = 10, health = 69, mana = Just 20 })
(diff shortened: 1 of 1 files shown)


ceddlyburge:
Hi Debanwita, I think the test is correct actually.

The instructions state that the damage done is double the mana cost, not that the mana cost is doubled.

The name of the test looks wrong though, how about we change it from 
`"Casting a spell spends double the mana"`
to 
`"Casting a spell causes damage of double the mana"'?

exercises/concept/role-playing-game/tests/Tests.elm
```diff
@@ -33,7 +33,7 @@ tests =
             [ test "Casting a spell spends double the mana" <|
```
ceddlyburge, line 33:
```suggestion
            [ test "Casting a spell causes damage of double the mana" <|
```

exercises/concept/role-playing-game/tests/Tests.elm
```diff
@@ -33,7 +33,7 @@ tests =
...
                 \() ->
                     Expect.equal (castSpell 9 { name = Nothing, level = 10, health = 69, mana = Just 20 })
-                        ( { name = Nothing, level = 10, health = 69, mana = Just 11 }, 18 )
+                        ( { name = Nothing, level = 10, health = 69, mana = Just 2 }, 18 )
```
ceddlyburge, line 36:
```suggestion
                        ( { name = Nothing, level = 10, health = 69, mana = Just 11 }, 18 )
```

ceddlyburge (changes requested):

mpizenberg:
Indeed, the test is correct, but the test wording is wrong. I like the new wording by @ceddlyburge better too.

debanwita27:
The initial wording made me misunderstand the case. Thanks for the explanation :D

ceddlyburge (approved):

ceddlyburge:
Yep, thanks for spotting the problem and bringing it to our attention :)

PR #577, opened 2023-06-07
Adds parameter to MonsterAttack.elm attack3

pwadsworth:
Missing parameter in attack3 was causing runtime exception.


pwadsworth:
... and it was for a reason

mpizenberg:
Indeed ^^ that’s the exercise related to partial application and composition so the goal is to write functions by composing other functions instead of explicitely handling the "unrolled" version with all parameters specified.

ceddlyburge:
Hi Pete, if the docs / instructions were confusing please let us know and we can have a think. This exercise is fairly new and its hard getting things right first time ...

PR #580, opened 2023-06-30
fix tests of concept exercise 'monster-attack'

DerTee:
Until now the tests passed even if the exercise functions didn't concatenate their content to the input string, because the tests always provided an empty input string.

### DIFF
--- a/exercises/concept/monster-attack/tests/Tests.elm
+++ b/exercises/concept/monster-attack/tests/Tests.elm
@@ -11,50 +11,50 @@ tests =
         [ describe "1"
             [ test "attackWithSword1 should update Monster with desired description" <|
                 \_ ->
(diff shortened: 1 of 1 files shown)


ceddlyburge (approved):
I opened a PR for this, after the chat on the forum, but nobody seems to want to approve it, so this is better, as I can approve it!

How about putting a space after the full stop? I've made that change, so if you're happy with it I'll merge this in.

DerTee:
> I opened a PR for this, after the chat on the forum, but nobody seems to want to approve it, so this is better, as I can approve it!
Oh, sorry about that! Maybe you can ask for a promotion and in the future get your changes merged without a review?

> How about putting a space after the full stop? I've made that change, so if you're happy with it I'll merge this in.
Of course, thank you for the fix 👍

Issue #536, opened 2022-11-27
Concept: Strings


jiegillet:
# Design

## Goal

The Goal is to learn to use the `String` and `Char` modules.

## Learning objectives

Students should be able to
- insert unicode characters in strings
- verify `Char` properties (`isAlpha`, `isUpper`...)
- use basic string functions (`length`, `replace`...)
- split and join strings (`split`, `words`, `lines`, `(++)`, `concat`, `join`...)
- cut up strings (`slice`, `left`...)
- check for substrings (`contains`, `startsWith`...)
- convert to and from other types (`fromInt`, `toInt`...)
- format a string (`toUpper`, `toLower`, `trim`...)
- use some higher order functions (`all`, `any`, `filter`...)

## Out of scope

- Nothing in particular, although there is probably too much to cover in details

## Concepts

The concept this exercise unlocks is:

- `strings`

## Prerequisites

- booleans
- maybe
- lists

## Analyzer

TBD depending on the exercise

pwadsworth:
Seems no one is working on this. If that is the case, I am open to take a stab at it.

ceddlyburge:
Hi Pete, that would be great, thanks!

pwadsworth:
Is the concept already written? If so, I would like to see it to make the exercise match its main topics. If not, I will write the concept as well.

mpizenberg:
I don't think we wrote anything yet. Are you familiar already with exercism repository structure?

pwadsworth:
Yes. I checked the concepts directory. There is no strings concept there.

PR #605, opened 2023-09-20
Adds strings concept

pwadsworth:
Adds concept files for Strings per Learning Objectives posted in Issue #536. Includes description of functions in `Char` and `String` modules.


concepts/strings/about.md
```diff
@@ -0,0 +1,220 @@
...
+"""
+    Code's strings entwine
+    , In binary, thoughts combine
+    , Logic's elegant line.
+"""
+```
+
+## String Manipulation
+
+`reverse : String -> String` reverses the string.
```
ceddlyburge, line 29: There is a convention in the repo to use `-->` to indicate the outcome of a function in the documentation like this.
To be fair, I don't think it is documented anywhere.

So maybe something like this:

Strings are reversed using `String.reverse`
```
reverse "Hello world." --> ".dlrow olleH"
```

You could have a look at one of the other concepts and try and follow the pattern. I think there is a pattern to the other concepts, but also, it probably isn't 100% followed, so let me know if you have a strong objection to it.

concepts/strings/about.md
```diff
@@ -0,0 +1,220 @@
...
+## Strings
+
+The `String` type provides a built-in representation for efficient string manipulation and can represent any sequence of unicode characters.
+Strings are represented with double quotes: `"this is a string"`.
+Multi-line strings are represented with triple double quotes and can have unescaped quotes and newlines.
+
+```elm 
+haiku = 
+"""
+    Code's strings entwine
```
ceddlyburge, line 21: I enjoyed this :)

ceddlyburge, line 21: I enjoyed this :)

concepts/strings/about.md
```diff
@@ -0,0 +1,220 @@
...
+## Getting Substrings
+
+`slice : Int -> Int -> String -> String` Takes a substring given a start and end index.
+Negative indexes are taken starting from the end of the list.
+```elm
+slice  0  6 "snakes on a plane!" == "snakes"
+```
+
+`left : Int -> String -> String` and `right : Int -> String -> String
+` take n characters from the left or right side of the string respectively.
```
ceddlyburge, line 95: This looks like a typo

concepts/strings/introduction.md
```diff
@@ -0,0 +1,220 @@
+# Introduction
```
ceddlyburge, line 1: The introduction is meant to be a bit more of a summary, so should be a bit shorter than about.md.

ceddlyburge (commented):
Hi Pete, this looks good. I really enjoyed all the examples!
I've left some comments, but nothing drastic, can you take a look?
Thanks, Cedd

concepts/strings/about.md
pwadsworth: Corrected.

concepts/strings/about.md
pwadsworth: No problem. I changed all the `==` to `-->' indented on the following line. Other concepts in this track vary between inline and indented on the next line. I think the latter improves readability and is more in keeping with Elm format rules.

concepts/strings/introduction.md
pwadsworth: I summarized the information in a table.

concepts/strings/about.md
```diff
@@ -0,0 +1,265 @@
...
+```elm
+append "butter" "fly" 
+    --> "butterfly"
+"butter" ++ "fly" 
+    --> "butterfly"
+```
+
+`concat : List String -> String` concatenates a list of strings into a single string.
+```elm
+concat ["never","the","less"]
```
ceddlyburge, line 58: I think it would be good to format all the arrays consistently.

Mostly there are no spaces (`["1","2"]`) in this document, but occasionally there are spaces (`["1", "2"]`).

I think it probably also makes sense to go with whatever elm-format does, which I think is `[ "1", "2" ]`.

Other than this I think the concept looks great.

ceddlyburge (approved):
Hi @mpizenberg / @jiegillet , this looks good to me, can one of you give it a look as well, in case I have missed anything.
Thanks, Cedd

concepts/strings/introduction.md
ceddlyburge: Awesome, thats a useful reference :)

concepts/strings/about.md
pwadsworth: Thanks for the feedback Cedd.

I updated the lists' format and the other recommendations. I think this is ready to merge.

What do you think of using the "99 Bottles of Beer" problem for this concept? I feel it is simple enough that the focus will be in the use of String. My only concern is that it requires recursion which is not included in the concepts yet, although it is used in other exercises.

mpizenberg:
Hi, I don't think I can take the time this weekend, next weekend is more
likely for me.

concepts/strings/about.md
ceddlyburge: Hi Pete, until now we have always created new exercises when we have made concepts, so I don't know what the repercussions would be of using an existing one. I think you would have to duplicate it as a concept exercise, which might be a bit strange. Maybe we could deprecate the old one or something.

Anyway, hopefully there is a better way. This page shows a list of concepts and their associated concept exercises across all the exercism tracks.
https://exercism.org/docs/building/tracks/stories

Quite a few tracks have already implemented a string concept, so it is probably easier to repurpose one / copy paste one. And that will hopefully make it easier to work in more of the string functions.

I'm going to wait for Matthieu or Jie to approve this before merging, which sounds like it might take a week, apologies for that. I don't think there will be many changes after their review though, so you could start on the concept exercise now if you want, probably in a new branch.

pwadsworth: Cedd, no problem waiting for the review. 

IRT using 99-Beers, I do not see it in the current Elm exercises. Is it by another name? I was not able to find anything similar. Maybe I worded the question poorly, what I meant was whether you thought that problem was well suited, not that it was already written.

I started working on 99 Beers and it seems a good fit to introduce the basics, but more exercises will be needed to touch more of the String module functionality. I will open a different issue and PR for that.

ceddlyburge: Ah I see. I've done that 99 beers problem in other languages so just assumed it was implemented in Elm, but obviously  not :)

concepts/strings/introduction.md
jiegillet: I think this is a bit too summarized, this is the entry point for the students, so I would at least show a string, a multiline string, a couple of common functions spanning a few different sections (concat, `++`, split, join, replace, length, toInt, contains, map, or something like that?)

jiegillet (commented):
This looks fantastic, thank you!
I left a bunch of small formatting changes/tiny suggestions, but the one change I would make is to add a few more examples in the intro.

Concerning the discussion on the concept exercise:
1. I don't think we should use a practice exercise as a concept exercise. They have slightly different roles, and usually the concept exercises have a low difficulty while showcasing as many of the aspects of the concept as possible (which usually means they are a bit contrived)
2. Don't implement `beer-song` it has recently been [deprecated](https://github.com/exercism/problem-specifications/blob/main/exercises/beer-song/.deprecated) in favor of `bottle-song` (which is virtually the same, just without alcohol).

pwadsworth:
@mpizenberg I committed all the edits and made the recommended changes to About.md and the examples in Introduction.md.

pwadsworth:
>     2. Don't implement `beer-song` it has recently been [deprecated](https://github.com/exercism/problem-specifications/blob/main/exercises/beer-song/.deprecated) in favor of `bottle-song` (which is virtually the same, just without alcohol).

That takes all the fun out of the song and kills the context story I was using. I will use the Squeaky Clean exercise then to touch on more of the String module's functionality.

jiegillet (approved):
Thanks for doing the changes, I'm very happy with the result.

And also thank you for your understanding about the beer bottle exercise.

Issue #606, opened 2023-09-22
Concept Exercise for Strings


pwadsworth:
## Develop concept exercise(s) to practice the Strings concept covering the following topics:
1. Insert unicode characters in strings
2. Verify Char properties (isAlpha, isUpper...)
3. Use basic string functions (length, replace...)
4. Split and join strings (split, words, lines, (++), concat, join...)
5. Cut up strings (slice, left...)
6. Check for substrings (contains, startsWith...)
7. Convert to and from other types (fromInt, toInt...)
8. Format a string (toUpper, toLower, trim...)
9. Use some higher order functions (all, any, filter...)

## Recommendations:
- [99-Beers][beers]: Covers 1, 3, 4 (partial), and 7 above.
- [Squeaky Clean][squeaky]: Covers 1, 2, 3, 4, 5, 6, and 8 above

[squeaky]: https://github.com/exercism/csharp/blob/main/exercises/concept/squeaky-clean/.docs/instructions.md
[beers]: http://inventwithpython.com/pythongently/exercise23/

jiegillet:
There is one last thing we should do on this issue before closing: go through existing practice exercises and add the "strings" concept to the relevant ones ("prerequisites" field for concept exercises and "prerequisites"/"practices" for practice exercises). 
We already have the (obsolete) "topics" on the exercises, which sometimes contains "strings" already. That should help.

pwadsworth:
> There is one last thing we should do on this issue before closing: go through existing practice exercises and add the "strings" concept to the relevant ones ("prerequisites" field for concept exercises and "prerequisites"/"practices" for practice exercises). We already have the (obsolete) "topics" on the exercises, which sometimes contains "strings". That should help.

I'll tackle this over the next week.

jiegillet:
Closed via #629

PR #607, opened 2023-09-22
Concept Exercise: Strings

pwadsworth:
Adds Beers exercise for the String Concept per Issue #606.

Covers: 
- Insert unicode characters in strings
- Use basic string functions (length, replace...)
- Split and join strings (split, words, lines, (++), concat, join...)
- Convert to and from other types (fromInt, toInt...)

### DIFF
--- a/exercises/concept/beers/.docs/instructions.md
+++ b/exercises/concept/beers/.docs/instructions.md
@@ -1,0 +1,46 @@
+# Instructions
+
+You are the new bosun of a small pirate crew that just killed their previous bosun for fowling their merriment with a bad song. As pirates would have it, your crew likes to pass the time getting loaded to the gunwales while crossing the briny deep and you are in charge of rationing the beer. The Cap'n has tasked you to minimize the number of resupply trips to make the best possible time between hauls. To achieve these conflicting requirements, and keep your crew from blowing you down when you are also loaded to the gunwales, you decide to stick to the tried-and-true 99 Bottles of Beer song using a program to help you keep the song straight and remind you to make a beer run when the ship is running low.
+
+## 1. Return the last stanza when there are 0 beers.
+
+Implement a `beerSongFrom` function that returns the last two verses of the beer song, in separate lines, when given the value `0`.
+
+> No more bottles of beer on the wall, no more bottles of beer.</br>
+> Go to shore to get some more, 99 bottles of beer on the wall!</br>
+
+## 2. Return the last two stanzas when starting with 1 beer.
+
+Refactor your `beerSongFrom` function to return the second to last stanza of the beer song when given the value `1`, followed by the rest of the song. Make sure your stanzas are properly separated by one empty line!
+
+> 1 bottle of beer on the wall, 1 bottle of beer.</br>
+> Take it down and pass it around, no more bottles of beer on the wall.<br><br>...
+    
+## 3. Return the last three stanzas when starting with 2 beers.
+
+Refactor your `beerSongFrom` function once again to return the third to last stanza when given the value `2`, followed by the rest of the song.
+> 2 bottles of beer on the wall, 2 bottles of beer.</br>
+> Take one down and pass it around, 1 bottle of beer on the wall.<br><br>...
+
+## 4. Return any number of verses with the correct ending.
+
+Your crew is never satisfied with less than 30 beers, and you realize your poor writing skills are likely to make you foul the song if you try to write every stanza by hand. So, you decide to refactor the function one last time to make it work for any number of beers. You also want the last stanza to change to show the crew's satisfaction when they had at least 30 beers, or remind you it is time to get more beer to keep them happy if you started from less than 30.
+> ... <br><br>
+> 1 bottle of beer on the wall, 1 bottle of beer.</br>
+> Take it down and pass it around, no more bottles of beer on the wall.
+
+Should be followed by either
+
+> No more bottles of beer on the wall, no more bottles of beer.</br>
+> Go to the shore to get some more, 99 bottles of beer on the wall!
+
+if they started with less than 30 beers, or
+
+> No more bottles of beer on the wall, no more bottles of beer.</br>
+> We've taken them down and passed them around; now we're all drunk and passed out!
+
+if they had 30 beers or more.
+
+## Present your song to the crew!
+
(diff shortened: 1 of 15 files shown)


pwadsworth:
I don't see why configlet is failing with:

```
Run configlet fmt
Found 21 Concept Exercises and 53 Practice Exercises in /home/runner/work/elm/elm/config.json
Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
or '.articles/config.json' file...
The below paths are relative to '/home/runner/work/elm/elm'
Not formatted: exercises/concept/beers/.meta/config.json
Error: Process completed with exit code 1.
```

The config.json looks the same as other exercises'.

exercises/concept/beers/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
+# Instructions
+
+You are the new bosun of a small pirate crew that just killed their previous bosun for fowling their merriment with a bad song. As pirates would have it, your crew likes to pass the time getting loaded to the gunwales while crossing the briny deep and you are in charge of rationing the beer. The Cap'n has tasked you to minimize the number of resupply trips to make the best possible time between hauls. To achieve these conflicting requirements, and keep your crew from blowing you down when you are also loaded to the gunwales, you decide to stick to the tried-and-true 99 Bottles of Beer song using a program to help you keep the song straight and remind you to make a beer run when the ship is running low.
```
ceddlyburge, line 3: This in a very entertaining introduction to the exercise, and I thoroughly enjoyed reading it. I think it will probably make difficult reading for a non native english speaker though. Probably @mpizenberg and @jiegillet can comment on this, as they are non native english speakers.
Maybe we can find a way to keep the fun text but also make life easier for students. We can have a think, but maybe just a bit at the start stating that the following paragraph is not critical for the exercise.

exercises/concept/beers/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+## 1. Return the last stanza when there are 0 beers.
+
+Implement a `beerSongFrom` function that returns the last two verses of the beer song, in separate lines, when given the value `0`.
+
+> No more bottles of beer on the wall, no more bottles of beer.</br>
+> Go to shore to get some more, 99 bottles of beer on the wall!</br>
+
+## 2. Return the last two stanzas when starting with 1 beer.
```
ceddlyburge, line 12: It's generally smoother for the student if later exercises don't change the values of earlier ones, and that once a part of the exercise is successfully completed, the tests for that part pass.

I think that asking the student to refactor the existing function will break this pattern, and mean that the tests will fail even when part 1 is done successfully, so then the student doesn't know whether they have completed part 1 or not.

Maybe this can be avoided by creating a `beerSongFrom0` in part 1, and a `beerSongFrom1` here?

exercises/concept/beers/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+Refactor your `beerSongFrom` function to return the second to last stanza of the beer song when given the value `1`, followed by the rest of the song. Make sure your stanzas are properly separated by one empty line!
+
+> 1 bottle of beer on the wall, 1 bottle of beer.</br>
+> Take it down and pass it around, no more bottles of beer on the wall.<br><br>...
+    
+## 3. Return the last three stanzas when starting with 2 beers.
```
ceddlyburge, line 19: Same here ...

exercises/concept/beers/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+> Take one down and pass it around, 1 bottle of beer on the wall.<br><br>...
+
+## 4. Return any number of verses with the correct ending.
+
+Your crew is never satisfied with less than 30 beers, and you realize your poor writing skills are likely to make you foul the song if you try to write every stanza by hand. So, you decide to refactor the function one last time to make it work for any number of beers. You also want the last stanza to change to show the crew's satisfaction when they had at least 30 beers, or remind you it is time to get more beer to keep them happy if you started from less than 30.
+> ... <br><br>
+> 1 bottle of beer on the wall, 1 bottle of beer.</br>
+> Take it down and pass it around, no more bottles of beer on the wall.
+
+Should be followed by either
```
ceddlyburge, line 32: Does this changing of the final verse teach any `string` concepts? It looks like it is more about conditionals and scope ...

exercises/concept/beers/tests/Tests.elm
```diff
@@ -0,0 +1,352 @@
...
+                        |> Expect.equal txtBeers.one
+        , skip <|
+            test "2 bottles" <|
+                \_ ->
+                    beerSongFrom 2
+                        |> Expect.equal txtBeers.two
+        , skip <|
+            test "99 bottles" <|
+                \_ ->
+                    beerSongFrom 99
```
ceddlyburge, line 28: I don't think there is a need for the function to test the 99 case, it makes this file unnecessarily long! Maybe the >30 case as that is special ...

ceddlyburge (commented):
Hi Pete. Overall this looks like a fun exercise to solve, but the exemplar solution only uses String.replace, String.fromInt and ++, which seems like a small amount of string functions to be using in an exercise all about strings!
Maybe the squeaky clean exercise would be a better fit?

ceddlyburge:
Also, if we do use a different exercise for strings, we could still use this PR, by modifying it a bit and turning it in to the standard 99 bottles exercise

pwadsworth:
Hi Cedd, I was thinking in a similar way.: 99 Bottles as a second exercise for strings and Squeaky Clean as the primary. 

That said, if 99 Bottles is not primary for Strings and doesnt block advancing on concepts, wouldn't it be better used to also practice/introduce refactoring? I was going for an approach similar to 99 Bottles of OOP (saw it on your blog and started reading it, and need to add the part I just got to this weekend 😁) driving the refactoring through the exercise steps. 

If that's out of scope for the track, it is an easy change to make a straightforward string replacement exercise. Let me know what you think.

pwadsworth:
Changing to Draft status until decision is made on whether to use or delete. Mention of alcohol not withstanding, I think this can be used for practicing/introducing "fearless refactoring." But, I understand it might be out of scope for the intent of the track.

ceddlyburge:
I think using this as a practive exercise is definitely a workable idea, but maybe we could park that for now. It is definitely better to have the tests for a task pass when the task is done, but I think that could be achieved by copy pasting each function to a new function before refactoring it. Also Green Bottles / 99 Bottles is a canonical exercise that already has its own tests and text and similar in the template repo, so we would probably want to make it clearly distinct from that. It's all doable though I think.

jiegillet:
Superseded by #611

PR #609, opened 2023-09-30
Add practice exercise `binary-search-tree`

jiegillet:
I've been away for a while, but I'm trying to get back in.
Practice exercises usually motivate me, so here is a nice little one :)

### DIFF
--- a/exercises/practice/binary-search-tree/.docs/instructions.md
+++ b/exercises/practice/binary-search-tree/.docs/instructions.md
@@ -1,0 +1,47 @@
+# Instructions
+
+Insert and search for numbers in a binary tree.
+
+When we need to represent sorted data, an array does not make a good data structure.
+
+Say we have the array `[1, 3, 4, 5]`, and we add 2 to it so it becomes `[1, 3, 4, 5, 2]`.
+Now we must sort the entire array again!
+We can improve on this by realizing that we only need to make space for the new item `[1, nil, 3, 4, 5]`, and then adding the item in the space we added.
+But this still requires us to shift many elements down by one.
+
+Binary Search Trees, however, can operate on sorted data much more efficiently.
+
+A binary search tree consists of a series of connected nodes.
+Each node contains a piece of data (e.g. the number 3), a variable named `left`, and a variable named `right`.
+The `left` and `right` variables point at `nil`, or other nodes.
+Since these other nodes in turn have other nodes beneath them, we say that the left and right variables are pointing at subtrees.
+All data in the left subtree is less than or equal to the current node's data, and all data in the right subtree is greater than the current node's data.
+
+For example, if we had a node containing the data 4, and we added the data 2, our tree would look like this:
+
+      4
+     /
+    2
+
+If we then added 6, it would look like this:
+
+      4
+     / \
+    2   6
+
+If we then added 3, it would look like this
+
+       4
+     /   \
+    2     6
+     \
+      3
+
+And if we then added 1, 5, and 7, it would look like this
+
+          4
+        /   \
+       /     \
+      2       6
+     / \     / \
+    1   3   5   7
--- a/exercises/practice/binary-search-tree/.meta/src/BinarySearchTree.example.elm
+++ b/exercises/practice/binary-search-tree/.meta/src/BinarySearchTree.example.elm
@@ -2,13 +2,13 @@ module BinarySearchTree exposing (BinaryTree(..), makeTree, sort)
 
 
 type BinaryTree
-    = Nil
+    = Empty
     | Tree BinaryTree Int BinaryTree
 
 
 makeTree : List Int -> BinaryTree
 makeTree =
-    List.foldl insert Nil
+    List.foldl insert Empty
 
 
 sort : List Int -> List Int
@@ -19,8 +19,8 @@ sort =
 insert : Int -> BinaryTree -> BinaryTree
 insert n tree =
     case tree of
-        Nil ->
-            Tree Nil n Nil
+        Empty ->
+            Tree Empty n Empty
 
         Tree left value right ->
             if n <= value then
@@ -33,7 +33,7 @@ insert n tree =
 toList : BinaryTree -> List Int
 toList tree =
     case tree of
-        Nil ->
+        Empty ->
             []
 
         Tree left value right ->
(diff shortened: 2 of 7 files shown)


exercises/practice/binary-search-tree/tests/Tests.elm
```diff
@@ -0,0 +1,65 @@
...
+
+import BinarySearchTree exposing (BinaryTree(..))
+import Expect
+import Test exposing (Test, describe, skip, test)
+
+
+tests : Test
+tests =
+    describe "BinarySearchTree"
+        [ -- skip <|
```
ceddlyburge, line 11: These skips need to be restored I think

ceddlyburge (approved):
This all looks good. Presumably the copy and tests are all standard and shared between the other languages so don't need to be examined in too much detail?

ceddlyburge:
@pwadsworth, do you want to give this a review? It seems like you are keen to get involved, and reviewing stuff is very useful for us, and a good way of getting familiar with the codebase and the (sometimes unwritten) conventions and things like that.

I remember Jie reviewing one of my PRs before he became a maintainer, and obviously we all review each others code all the time.

exercises/practice/binary-search-tree/tests/Tests.elm
jiegillet: Oops! Done in [e0a1763](https://github.com/exercism/elm/pull/609/commits/e0a1763ab98bbddf147416473a790c715834d324)

jiegillet:
> This all looks good. Presumably the copy and tests are all standard and shared between the other languages so don't need to be examined in too much detail?

Mostly, but I changed the function names, I thought the canonical ones weren't very good: `data` -> `makeTree` and `sortedData` -> `sort`.

ceddlyburge:
Welcome back! I agree about the canonical names. Is it worth raising a PR in the repo to update them for everybody?

pwadsworth:
> @pwadsworth, do you want to give this a review? It seems like you are keen to get involved, and reviewing stuff is very useful for us, and a good way of getting familiar with the codebase and the (sometimes unwritten) conventions and things like that.

Love to. I should have some time to go over it later today.

exercises/practice/binary-search-tree/.docs/instructions.md
```diff
@@ -0,0 +1,47 @@
+# Instructions
+
+Insert and search for numbers in a binary tree.
+
+When we need to represent sorted data, an array does not make a good data structure.
+
+Say we have the array `[1, 3, 4, 5]`, and we add 2 to it so it becomes `[1, 3, 4, 5, 2]`.
+Now we must sort the entire array again!
+We can improve on this by realizing that we only need to make space for the new item `[1, nil, 3, 4, 5]`, and then adding the item in the space we added.
```
pwadsworth, line 9: I think this needs to be adapted to Elm to avoid the invalid code (list with multiple types), and the use of nil since a big part of Elm is the deliberate effort not to use it.

exercises/practice/binary-search-tree/.meta/src/BinarySearchTree.example.elm
```diff
@@ -0,0 +1,40 @@
+module BinarySearchTree exposing (BinaryTree(..), makeTree, sort)
+
+
+type BinaryTree
+    = Nil
```
pwadsworth, line 5: Recommend renaming `Nil` --> `Empty` to avoid confusion with Elm not supporting the concept of nil.

exercises/practice/binary-search-tree/src/BinarySearchTree.elm
```diff
@@ -0,0 +1,16 @@
+module BinarySearchTree exposing (BinaryTree(..), makeTree, sort)
+
+
+type BinaryTree
+    = Nil
```
pwadsworth, line 5: Recommend renaming `Nil` --> `Empty` to avoid confusion with Elm not supporting the concept of nil.

pwadsworth (commented):
This looks pretty good to me. My only recommendations are to adapt some of the terms to match Elm concept of no null values and immutability i.e. using the term parameter/argument instead of variable.

exercises/practice/binary-search-tree/.docs/instructions.md
ceddlyburge: This is a good point I think, and happens because this copy comes from the canonical source. I'm not even sure that line 9 adds much anyway, so maybe we could just remove it.

exercises/practice/binary-search-tree/.meta/src/BinarySearchTree.example.elm
ceddlyburge: I think Nil in a custom type is fairly idiomatic in Elm, although I'm not quite sure. Nothing is used within Maybe, and other modules tend to use something else to avoid a name clash as Maybe is so common, and I think Nil is often what people go for.

pwadsworth: > ...Nothing is used within Maybe, and other modules tend to use something else to avoid a name clash as Maybe is so common, and I think Nil is often what people go for.

I was recommending a custom type 'Empty' for that reason. Either way, I don't think it's a significant point, just a more idiomatic option.

exercises/practice/binary-search-tree/.docs/instructions.md
jiegillet: We can't modify the instructions, they are standard for all tracks. At best, we can add an addendum, but I don't think it's worth it, many exercises are very vague, and in Exercism, tests are king.

exercises/practice/binary-search-tree/.meta/src/BinarySearchTree.example.elm
jiegillet: I also thought `Nil` was quite idiomatic in Elm, but most tutorials I could find in my extensive 5-minute search reported using `Empty`. Maybe it's more common in Haskell?
Anyway, made the change in [d09f956](https://github.com/exercism/elm/pull/609/commits/d09f956382f5cc6e15b885273b637363fd2db751)

ceddlyburge:
Hi @jiegillet , I think this all looks good, do you want to merge it whenever you are happy?

PR #612, opened 2023-10-13
Fix magician in training introduction

ThunderKey:
* Fixed the import example that requires parentheses
* Fixed link to default imports

### DIFF
--- a/concepts/array/about.md
+++ b/concepts/array/about.md
@@ -18,7 +18,7 @@ Array Int --> an array of Int
 Arrays can be defined as follows:
 
 ```elm
-import Array exposing Array
+import Array exposing (Array)
 
 empty : Array Char
 empty = Array.empty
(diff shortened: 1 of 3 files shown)


jiegillet:
That's awesome, thank you!

However, that file is a copy of `concepts/array/introduction.md`, so if you could change the original, that would be perfect. Please also check `concepts/array/about.md`, it's a very similar file and it's likely that it contains the same mistakes.
Once you did this, please run `bin/fetch-configlet` then `bin/configlet generate`. The first command gets configlet (which is a tool to help with various exercism tasks) and the second one copies the introduction files (it knows how to do that thanks to `exercises/concept/magician-in-training/.docs/introduction.md.tpl`).

ThunderKey:
That's an awesome setup you have! Thanks for the quick introduction.

I've taken the liberty to force-push the changes directly.

jiegillet (approved):
Thank you, much appreciated :)

PR #617, opened 2023-10-19
Add practice exercise `pop-count`

### LINKED ISSUE #615: Implement new pop-count
I'd like to port this exercise introduced in exercism/problem-specifications#2333.

BNAndras:
Will close #615

### DIFF
--- a/config.json
+++ b/config.json
@@ -580,8 +580,13 @@
         "slug": "pop-count",
         "name": "Pop Count",
         "uuid": "058c69bd-9c1a-45f0-a94f-62be9f0c9e4e",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "basics-2"
+        ],
+        "prerequisites": [
+          "basics-1",
+          "basics-2"
+        ],
         "difficulty": 2
       },  
       {
--- a/config.json
+++ b/config.json
@@ -577,6 +577,14 @@
         ]
       },
       {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "058c69bd-9c1a-45f0-a94f-62be9f0c9e4e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
+      },  
+      {
         "slug": "raindrops",
         "name": "Raindrops",
         "uuid": "e3e73171-c3d1-49f7-9511-d1f4a08e7a69",
--- a/config.json
+++ b/config.json
@@ -2,6 +2,12 @@
   "language": "Elm",
   "slug": "elm",
   "active": true,
+  "status": {
+    "concept_exercises": true,
+    "test_runner": true,
+    "representer": true,
+    "analyzer": true
+  },
   "blurb": "Elm is a friendly functional language for the Web",
   "version": 3,
   "online_editor": {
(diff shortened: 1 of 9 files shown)


BNAndras:
Oops, I had elm-review suggest pruning unneeded dependencies from elm.json and didn't think twice about it.

ceddlyburge:
> Oops, I had elm-review suggest pruning unneeded dependencies from elm.json and didn't think twice about it.

Yeah, this is a bi of an annoyance. All the elm.json files must be the same, its a requirement of the test runner, which has to makle sure that it has all the dependencies in the elm.json files.

config.json
```diff
@@ -576,6 +576,14 @@
           "sorting"
         ]
       },
+      {
+        "slug": "pop-count",
+        "name": "Pop Count",
+        "uuid": "058c69bd-9c1a-45f0-a94f-62be9f0c9e4e",
+        "practices": [],
```
ceddlyburge, line 583: Are there any concepts this exercise practices, or that are required?

ceddlyburge (commented):
This all looks good to me. I'll let Jie / Matthieu take a look as well, and you'll need to fix up the elm.json in order for the build to pass. I'll approve when it does. The tests can be flaky, so if that happens let me know and I'll re run them.
Thanks for the contribution!
Cheers, Cedd

config.json
BNAndras: I wasn't sure. Normally, I'd gravitate towards a `numbers` concept if available for both the practices or requires arrays. Since that's not here, perhaps `basics-2` for both since it touches upon numbers?

jiegillet: Yes, I think `basics-2` would be great for the practices fiels, and `basics-1`, basics-2 for the prerequisites. It's nice to have an exercise available so early in the tree.

jiegillet (commented):
I'm good with this as well (once the things already discussed are settled), thank you for porting this.

I was wondering if we needed an analyzer comment to prevent people from using a built-in `popCount`, but there is no such function in `Bitwise` and I think `elm/bytes` is also pretty unrelated, so we should be good.

config.json
```diff
@@ -2,6 +2,12 @@
   "language": "Elm",
   "slug": "elm",
   "active": true,
+  "status": {
+    "concept_exercises": true,
+    "test_runner": true,
+    "representer": true,
+    "analyzer": true
+  },
```
jiegillet, line 10: That's too many changes on the config, I can't be sure of what the actual change is. I'm assuming it got sorted somehow, but could you revert that?

BNAndras: I had to run configlet fmt to make the CI pass. There was a recent release for configlet so I think that’s the issue. I can open a separate PR for the config formatting. That’ll make it easier to see the actual changes.

jiegillet: Oh I see, thank you for pointing it out. Please leave it like that, I checked that it's only adding what it's supposed to.

But that's strange, when I run `bin/configlet fmt` is says everything is clear, but running `bin/configlet fmt --update` on main it changes this file. Is it the same for you? I should flag that.

jiegillet (approved):
Thank you for adding this one :)

PR #618, opened 2023-10-21
Prevent some stubs from making the tests crash

jiegillet:
Usually test runner handles the `Debug.todo` of the stub files well, the tests fail but nothing else.
In `dnd-character`, the test runner crashes, because of the generators. 

I'm pretty sure it's not because of fuzz tests, because it also crashes without them as long as you're trying to use the generators (with `Random.step` for example).

After playing around, I found that wrapping the `Debug.todo` in a `Random.lazy` fixes the issue. It makes it slightly more messy for the students, but it's much better than crashing, because as of now they have to complete all of the functions before running any test.

@mpizenberg do you think I should open an issue about this? What would the right place be? `elm-explorations/test`? `rtfeldman/node-test-runner`?  `elm-test-rs`? (edit: I've found that the tests are really finicky with what crashes and what doesn't, maybe it's not worth it).

After playing around more, I found more exercises with a stub that makes the test runner crash, it's kind of finicky. In one case, it was a variable name shadowing, in another the functions needed to have the arguments, I fixed those.

There are still 3 exercises that make the test runner crash: `custom-set`, `grade-school` and `zebra-puzzle`. I've decided not to address these:
- for `custom-set` and `grade-school` the problem is the first function: `empty` (for some reason). But since this is the very first function the student would create, it's not much of an issue.
- for zebra-puzzle`, the problem may be that `drinksWater` and `ownsZebra` are constants? I don't know. But there are no real tests in that exercise anyway, so it doesn't matter

Also lots of concept exercises don't compile, but that's on purpose because the types need to be defined.

### DIFF
--- a/exercises/practice/binary-search-tree/src/BinarySearchTree.elm
+++ b/exercises/practice/binary-search-tree/src/BinarySearchTree.elm
@@ -7,10 +7,10 @@ type BinaryTree
 
 
 makeTree : List Int -> BinaryTree
-makeTree =
+makeTree data =
     Debug.todo "Please implement makeTree"
 
 
 sort : List Int -> List Int
-sort =
+sort data =
     Debug.todo "Please implement sort"
--- a/exercises/practice/complex-numbers/src/ComplexNumbers.elm
+++ b/exercises/practice/complex-numbers/src/ComplexNumbers.elm
@@ -24,7 +24,7 @@ fromPair pair =
 
 
 fromReal : Float -> Complex
-fromReal real =
+fromReal float =
     Debug.todo "Please implement fromReal."
 
 
--- a/exercises/practice/dnd-character/src/DndCharacter.elm
+++ b/exercises/practice/dnd-character/src/DndCharacter.elm
@@ -21,9 +21,9 @@ modifier score =
 
 ability : Generator Int
 ability =
-    Debug.todo "Please implement ability"
+    Random.lazy (\_ -> Debug.todo "Please implement ability")
 
 
 character : Generator Character
 character =
-    Debug.todo "Please implement character"
+    Random.lazy (\_ -> Debug.todo "Please implement character")


mpizenberg:
`Debug.todo` is compiled into a JS function that crashes on purpose. If it's used in a constant (a function without argument) the JS code generated for the constant contains a crash. So by definition, when the file is evaluated, it crashes right away. That is way we should not use `Debug.todo` directly without adding a parameter to a function, usually.

However there are some edge cases. If it's the first function the student has to implement, it's fine, not so much a big deal if it crashes. If the function must be written in a point-free style, we cannot add a parameter to the stub, as we ask the student to write a function with function compositions only (I see you added a parameter in some of these cases, make sure it's not asked of the student to write point-free style functions).

Regarding crashes with generators, it's possible that the code crashes on too many iterations with big data types being generated (list of list and that sort of things). This behavior has been changed in the latest release of elm-exploration/tests, to let it run until it crashes with stack overflows or terminate normally. (Read Martin's post for more details).

After reading this, if there are still unexplained crashes, can you tell us which ones please?

jiegillet:
> `Debug.todo` is compiled into a JS function that crashes on purpose. If it's used in a constant (a function without argument) the JS code generated for the constant contains a crash. So by definition, when the file is evaluated, it crashes right away. That is why we should not use `Debug.todo` directly without adding a parameter to a function, usually.

I see, that makes sense. I thought the function would need to be called before it crashed, but being (perceived as) a constant at the top level is enough to make it crash when the file loads.
 
> If the function must be written in a point-free style, we cannot add a parameter to the stub, as we ask the student to write a function with function compositions only (I see you added a parameter in some of these cases, make sure it's not asked of the student to write point-free style functions).

All good, I'm not aware of any practice exercise where we ask the students to use point-free style. Maybe a concept one, but I didn't touch those.

> Regarding crashes with generators, it's possible that the code crashes on too many iterations with big data types being generated (list of list and that sort of things). This behavior has been changed in the latest release of elm-exploration/tests, to let it run until it crashes with stack overflows or terminate normally. (Read Martin's post for more details).

That's not the case here, we don't have any generators defined in any stub file.

> After reading this, if there are still unexplained crashes, can you tell us which ones please?

No, that covers it. The crashing ones were all top level "constants".
Thanks for the input

mpizenberg (approved):

PR #613, opened 2023-10-14
Add concept `random` and concept exercise `maze-maker`

jiegillet:
Closes #537 

I had been thinking about doing this one for a while.
~~This is actually blocked by #549 (first time using fuzz tests) and https://github.com/exercism/elm-test-runner/pull/49 (parsing fuzztests)~~, but since I think there won't be big problems there and because I was motivated, I started working on this.

@pwadsworth if you are so inclined, I would love a review from you. If you could try to solve the exercise without looking at the exemplar and share your solution and impressions (is it too hard? too easy?) that would be great. As well as any other comment you might have of course.

The CI is expected to fail for two reasons: ~~the test-runner will not be able to parse the exercise~~ (EDIT: that's taken care of), and the introduction is missing from the exercise, I will copy it once it's been reviewed.

PR #619, opened 2023-10-21
Add practice exercise `simple-cipher`

jiegillet:
This is built on top of #613 so I can mention it practices the "random" concept.

This is the last practice exercise that was tagged with "random" on the problem specs. There is technically one more, but it's deprecated.

The instructions are not very good on this one, they mention step 1 and 2 and 3, but really only the last step needs to be implemented. However as usual we can't change touch them.


exercises/practice/simple-cipher/tests/Tests.elm
```diff
@@ -0,0 +1,140 @@
...
+                        |> Fuzz.listOfLengthBetween 1 100
+                        |> Fuzz.map String.fromList
+                    )
+                    "encoding then decoding gives the original message"
+                <|
+                    \message key ->
+                        message
+                            |> SimpleCipher.encode key
+                            |> SimpleCipher.decode key
+                            |> Expect.equal message
```
jiegillet, line 41: @ceddlyburge this one is a bit verbose because I had to generate string fuzzers with only `a` to `z`, but it's the perfect use of a fuzzer. The error message looks like
```
↓ SimpleCipher
↓ Substitution cipher
✗ encoding then decoding gives the original message

Given ("a","b")

    "c"
    ╷
    │ Expect.equal
    ╵
    "a"
```

exercises/practice/simple-cipher/.meta/src/SimpleCipher.example.elm
```diff
@@ -0,0 +1,58 @@
...
+encode =
+    transform Encode
+
+
+decode : String -> String -> String
+decode =
+    transform Decode
+
+
+transform : Direction -> String -> String -> String
```
ceddlyburge, line 21: I would probably separate this in to two functions (encode and decode) and have helper functions for the shared bits. It's a minor point though and I wouldn't be offended if you disagree.

Probably something like this

```
decode : String -> String -> String
decode = key text
    List.map2 decodeCombine (String.toList text) (longKey text)
        |> String.fromList
```

exercises/practice/simple-cipher/tests/Tests.elm
ceddlyburge: I agree, this looks great. Maybe create a new fuzzer that generates the key? It's used in enough places to justify it I think.

ceddlyburge (approved):

exercises/practice/simple-cipher/.meta/src/SimpleCipher.example.elm
jiegillet: Well, I already had encode and decode, and this was the helper function, but I made some changes in [b6e57ae](https://github.com/exercism/elm/pull/619/commits/b6e57ae20cc945af2fd932f712989657d39a11a2), it it more clear?

ceddlyburge: Yep, that looks good. Maybe encodeHelper is the wrong name though, and the original transform name is better?

PR #620, opened 2023-10-21
Update casing for partial-application-composition name

BNAndras:
The other concepts with multiple words in the name are in Capitalized Case like "Pattern Matching", "Opaque Types", "Phantom Types", and so on. As a result, this concept sticks out weirdly when viewing the syllabus.


jiegillet (approved):
Makes Sense To Me, Thank You!

PR #621, opened 2023-10-23
Add practice exercise `rest-api`

jiegillet:
I'm now thinking about the `json` concept, and this is the most suitable practice exercise for it, maybe even the suitable one? I'm sure we could engineer some other to use JSON as raw data, I don't know.

I quite like my solution, it's pretty clean, but I don't really know about the general API. The description in problem-specs itself is pretty vague about it. Right now what I have is
```elm
  case RestApi.buildDatabase "{\"users\":[]}" of
      Err error ->
          Expect.fail ("== Database payload could not be parsed ==\n" ++ Decode.errorToString error)

      Ok database ->
          RestApi.post database "/add" "{\"user\":\"Adam\"}"
              |> Expect.equal "{\"name\":\"Adam\",\"owes\":{},\"owed_by\":{},\"balance\":0}"
```
I'm feeding the raw JSON into `buildDatabase` in one step, check if that worked, and then use `post`. I did this so that the parsing for the database could be done separately. 
Ideally the "server" would keep the database in an internal state, but it can't work in Elm.
Anyway, I'm happy for some feedback on that API.

This will require an analyzer check to encourage students to use `Json.Decode` and `Json.Encode`, because technically they could do it without.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1111,6 +1111,21 @@
           "random"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "rest-api",
+        "name": "REST API",
+        "uuid": "98b75220-b8bb-488b-9339-a3972f040c21",
+        "practices": [],
+        "prerequisites": [
+          "lists",
+          "comparison",
+          "dict",
+          "records",
+          "maybe",
+          "result"
+        ],
+        "difficulty": 8
       }
     ]
   },
--- a/exercises/practice/rest-api/.meta/src/RestApi.example.elm
+++ b/exercises/practice/rest-api/.meta/src/RestApi.example.elm
@@ -32,12 +32,12 @@ type PostPayload
 
 buildDatabase : String -> Result Error Database
 buildDatabase =
-    Decode.decodeString databaseD
+    Decode.decodeString databaseDecoder
 
 
 get : Database -> String -> Maybe String -> String
 get database url maybePayload =
-    case ( url, Maybe.map (Decode.decodeString getPayloadD) maybePayload ) of
+    case ( url, Maybe.map (Decode.decodeString getPayloadDecoder) maybePayload ) of
         ( "/users", Nothing ) ->
             Encode.encode 0 (encodeDatabase database)
 
@@ -53,7 +53,7 @@ get database url maybePayload =
 
 post : Database -> String -> String -> String
 post database url payload =
-    case ( url, Decode.decodeString postPayloadD payload ) of
+    case ( url, Decode.decodeString postPayloadDecoder payload ) of
         ( "/add", Ok (NewUser user) ) ->
             Encode.encode 0 (encodeUser user)
 
@@ -121,9 +121,9 @@ addIOU { lender, borrower, amount } database =
 -- DECODERS
 
 
-databaseD : Decoder Database
-databaseD =
-    userD
+databaseDecoder : Decoder Database
+databaseDecoder =
+    userDecoder
         |> Decode.list
         |> Decode.field "users"
         |> Decode.map
@@ -132,8 +132,8 @@ databaseD =
             )
 
 
-userD : Decoder User
-userD =
+userDecoder : Decoder User
+userDecoder =
     Decode.map4 User
         (Decode.field "name" Decode.string)
         (Decode.field "owes" (Decode.dict Decode.int))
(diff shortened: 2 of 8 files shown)


config.json
```diff
@@ -1111,6 +1111,21 @@
           "random"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "rest-api",
+        "name": "REST API",
+        "uuid": "98b75220-b8bb-488b-9339-a3972f040c21",
+        "practices": [],
```
jiegillet, line 1119: This will practice `"json"` when it exists

exercises/practice/rest-api/.meta/src/RestApi.example.elm
```diff
@@ -0,0 +1,187 @@
...
+
+
+type PostPayload
+    = NewUser User
+    | NewIOU IOU
+
+
+buildDatabase : String -> Result Error Database
+buildDatabase =
+    Decode.decodeString databaseD
```
ceddlyburge, line 35: I would probably call this databaseDecoder, but feels like a personal preference type of thing.

exercises/practice/rest-api/src/RestApi.elm
```diff
@@ -0,0 +1,36 @@
...
+    , owedBy : Dict Name Int
+    , balance : Int
+    }
+
+
+type alias Database =
+    Dict Name User
+
+
+buildDatabase : String -> Result Error Database
```
ceddlyburge, line 24: This is a difficult exercise, and if we can't change the instructions, then I think it would be worth adding some more help in this file (there is nothing in the instructions detailing what this function should do). I'm not exactly sure how best to do this, but maybe this function could be called databaseFromJsonString or something like that? Or we could supply this function ready made and ask the student to write the decoder, and then the type signature communicates what they need to do.

exercises/practice/rest-api/src/RestApi.elm
```diff
@@ -0,0 +1,36 @@
...
+buildDatabase payload =
+    Debug.todo "Please implement buildDatabase"
+
+
+get : Database -> String -> Maybe String -> String
```
ceddlyburge, line 29: I think a similar thing applies here. I don't think its totally clear that the return value should be a json string encoding the "Response with payload" from the instructions.

exercises/practice/rest-api/src/RestApi.elm
```diff
@@ -0,0 +1,36 @@
...
+get database url maybePayload =
+    Debug.todo "Please implement get"
+
+
+post : Database -> String -> String -> String
+post database url payload =
```
ceddlyburge, line 35: This is a bit easier to work out, as there is only one response in the instructions table, but in terms of making the exercise more approachable (and more about working with JSON, and less about working out what the exercise is asking) it also might be worth making things easier.

ceddlyburge (approved):
Hey Jie, you really have found a lot of enthusiasm, you're creating PRs faster than I can review them!
This exercise looks good, but is complicated. The complexity around encoding and decoding JSON is what we want the student to tackle, but there is also a lot of accidental complexity around understanding the exercise and what is required, and I think we could probably improve that. Probably by adding a bit more scaffolding to the initial code, or maybe by adding comments / more information in the Debug.todo's.

exercises/practice/rest-api/src/RestApi.elm
jiegillet: Maybe I could require that they return a `Json.Encode.Value`? 
Or more simply we could have a `type alias JsonString = String`.

ceddlyburge: Yeah something like that sounds good. I think we also need to state what the format of the json should be, there are two options in the table in the instructions, and we want the one from the "Response with payload" one. Maybe we could even define a type alias for it, then ask for a function to encode that, and then ask to compose that function here? I'm not quite sure on the details ...

jiegillet: I added the alias.
I thought that `Maybe JsonString` encapsulated well enough that the payload can missing, so I left it this way.

exercises/practice/rest-api/src/RestApi.elm
jiegillet: I thought about this a bunch, and I don't really want to simplify the exercise so that it's only about JSON decoding. That sounds like something a concept exercise should be doing, so I will leave this one the way it is and start thinking about a good concept exercise.

It's unfortunate that this one is really the only suitable practice exercise for JSON decoding but it's understandable. Most languages will use some kind of library and not think about it, Elm really stands out in that aspect, which is why this is one of the hardest concepts for newcomers.

PR #623, opened 2023-10-25
Disable representer

iHiD:
The elm representer is 500ing. This is now critical as it stops community solutions from showing. I'm disabling this and rolling back to the general default representer for this track.

### DIFF
--- a/config.json
+++ b/config.json
@@ -5,7 +5,7 @@
   "status": {
     "concept_exercises": true,
     "test_runner": true,
-    "representer": true,
+    "representer": false,
     "analyzer": true
   },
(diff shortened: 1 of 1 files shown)


ceddlyburge:
This is strange, as I don't think there has been a release of the Elm Representer for a long time ...

iHiD:
Yeah, I'm not sure. I imagine there's something more fundamentally wrong. I think @ErikSchierboom is going to take a look.

ceddlyburge:
Let me know if there is anything we can do ...

ErikSchierboom:
https://github.com/exercism/elm-representer/pull/28 should fix it

PR #626, opened 2023-10-29
Remove mention about no having analyzer check for `treasure-factory`

jiegillet:
To be merged after https://github.com/exercism/elm-analyzer/pull/63/files.
BTW thank you @ceddlyburge for dealing with this issue when I wasn't active.


exercises/concept/treasure-factory/.docs/instructions.md
```diff
@@ -36,7 +36,6 @@ Edit the type signatures of `secureChest` and `uniqueTreasures` to add the const
 ~~~~exercism/note
 Elm tests have access to exposed functions, but not to type signatures, it is therefore impossible for tests to verify that you are using the right signatures.
 Of course, the strongest indication that you have the right signatures is to succeed in getting the module and tests to compile and run.
```
ceddlyburge, line 38: Maybe we could add something here about what does happen (an analyser comment).

Something like 

```suggestion
Instead the analyser checks that the type signatures are correct, and will show errors if there is a problem.
```

ceddlyburge (approved):

PR #633, opened 2023-11-11
Run analyzer rules on all of the exercises in the CI

jiegillet:
Part of https://github.com/exercism/elm-analyzer/issues/65

The main idea is to make sure that exemplar still adhere to analyzer rules if they are modified.
While we are at it, I'm checking example solutions too.

### DIFF
--- a/exercises/practice/sgf-parsing/tests/Tests.elm
+++ b/exercises/practice/sgf-parsing/tests/Tests.elm
@@ -24,11 +24,17 @@ tests =
                 \() ->
                     SgfParsing.parse ";"
                         |> Expect.equal (Err "tree missing")
+
+        -- this test is structured a bit differently to avoid an elm-review NoUnused.CustomTypeConstructorArgs error
         , skip <|
             test "node without properties" <|
                 \() ->
-                    SgfParsing.parse "(;)"
-                        |> Expect.equal (Ok (Node { properties = Dict.empty, children = [] }))
+                    case SgfParsing.parse "(;)" of
+                        Ok (Node node) ->
+                            node |> Expect.equal { properties = Dict.empty, children = [] }
+
+                        _ ->
+                            Expect.fail "input didn't parse"
         , skip <|
             test "single node tree" <|
                 \() ->
--- a/exercises/practice/twelve-days/.meta/src/TwelveDays.example.elm
+++ b/exercises/practice/twelve-days/.meta/src/TwelveDays.example.elm
@@ -1,16 +1,9 @@
 module TwelveDays exposing (recite)
 
-import List exposing (map, range)
-import String exposing (concat)
-
 
 recite : Int -> Int -> List String
 recite start stop =
-    if withinSong start stop then
-        map lyric (range start stop)
-
-    else
-        Debug.todo "I would make up verses if I could...sorry about this"
+    List.map lyric (List.range start stop)
 
 
 lyric : Int -> String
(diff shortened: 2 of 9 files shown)


exercises/practice/sgf-parsing/tests/Tests.elm
```diff
@@ -24,11 +24,17 @@ tests =
                 \() ->
                     SgfParsing.parse ";"
                         |> Expect.equal (Err "tree missing")
+
+        -- this test is structured a bit differently to avoid an elm-review NoUnused.CustomTypeConstructorArgs error
```
jiegillet, line 28: I'm not suuuuper happy with this change, but it's better than ignoring the rule altogether for this exercise. For info, here is the full error:

```
-- ELM-REVIEW ERROR ------------------------------------ src/SgfParsing.elm:8:12

NoUnused.CustomTypeConstructorArgs: Argument is never extracted and therefore
never used.

7| type Tree
8|     = Node { properties : Dict String (List String), children : List Tree }
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

This argument is never used. You should either use it somewhere, or remove it at
the location I pointed at.
```

In a real project, that would highlight that no one is actually using this data structure, which is fair.

jiegillet:
Runs in 7 minutes. That's quite slow. We could speed it up a lot by doing the usual trick of putting all the files in the same folder, but this is more "realistic". Maybe it's not worth it.

jiegillet:
25 seconds. Much better. I will now try to commit bad code on purpose to see the CI failing.

jiegillet:
Failed as expected with the expected output:
```
{
  "summary": "Check the comments for things to fix. 🛠 ",
  "comments": [
    {
      "comment": "elm.strain.do_not_use_filter",
      "type": "essential",
      "params": {}
    },
    {
      "comment": "elm.treasure-factory.do_not_change_given_signatures",
      "type": "essential",
      "params": {}
    },
    {
      "comment": "elm.common.no_debug",
      "type": "actionable",
      "params": {}
    },
    {
      "comment": "elm.feedback_request",
      "type": "informative",
      "params": {}
    }
  ]
}
Error: Process completed with exit code 1.
```

exercises/practice/sgf-parsing/tests/Tests.elm
ceddlyburge: I think this is fine.

exercises/practice/twelve-days/.meta/src/TwelveDays.example.elm
```diff
@@ -1,16 +1,9 @@
...
-
 
 recite : Int -> Int -> List String
 recite start stop =
-    if withinSong start stop then
-        map lyric (range start stop)
-
-    else
-        Debug.todo "I would make up verses if I could...sorry about this"
+    List.map lyric (List.range start stop)
```
ceddlyburge, line 6: There is nothing in the instructions about what to do with invalid input. Maybe we should open a PR on the problem specifications repo?

ceddlyburge (approved):

exercises/practice/twelve-days/.meta/src/TwelveDays.example.elm
jiegillet: I don't think it's necessary, there are other exercises that practice input validation, this is a simple one early in the lreaning tree.

Issue #634, opened 2023-11-13
Exercise Binary Search Tree Missing Instructions and Hints


pwadsworth:
The current docs for Binary Search Tree only has an instructions.md file with content more adequate for an introduction.md. There are no specific instructions to what the student needs to accomplish, and no hints.md file for the exercise,

PR #635, opened 2023-11-13
Adds Instructions and Hints to Binary Search Tree

pwadsworth:
## Adds Instructions and Hints to Binary Search Tree.

- Added hints.md and introduction.md.
- Moved the content previously under instructions.md to introduction.md and replaced it with exercise instructions.

Answers issue #634

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ name: elm / pr
 on:
   pull_request:
   push:
-    branches: [ main ]
+    branches: [main]
 
 jobs:
   ci:
@@ -40,3 +40,13 @@ jobs:
 
       - name: Run tests for all exercises
         run: bin/run-all-exercises-in-docker.sh
+
+  elm-analyzer:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout exercism/elm
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
+
+      - name: Run tests for all exercises
+        run: bin/analyze-all-exercises-in-docker.sh
--- a/bin/analyze-all-exercises-in-docker.sh
+++ b/bin/analyze-all-exercises-in-docker.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e # Make script exit when a command fail.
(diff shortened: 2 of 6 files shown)


ceddlyburge:
Hi Pete, thanks for this!
There are some nuances / annoyances here I'm afraid.

I don't think hints.md is used for practice exercises (but will need to check). Certainly no other practice exercises currently  have them.

And there is a [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/binary-search-tree) repo, that contains the copy (and some other stuff) for all of the shared exercises.

This is where the copy comes from for this practice exercise. configlet is used to update this repo if / when things change in the problem specifications repo.

So basically some of this PR (and probably associated issue) need to be opened on the problem specifications repo.

I can review it there, but I don't have admin permissions on that, and the amount of thinking people do before merging things on that repo is quite high (as it affects all the tracks on exercism, and not just one). So we can add the introduction.md file here I think, but not the instructions.md.

Does that make sense? Sorry for the rambly language, I haven't used it much myself to be honest. I mostly work on concept exercises, which are specific to each track and don't come with this complication ...

Also, there is only one documentation file if the problem specifications repo (description.md), which relates to the instructions.md file here I think.

pwadsworth:
Hi Cedd, 

Sorry I didn't look at the other practice exercises. IRT instructions, the exercises from the problem-specification repo all tend to be lacking on instructions, making them difficult to understand what the intent of the exercise is. For many of them I have to go into the tests file to figure out what is expected from the different functions. I will read the problem-specification repo and open the issue there if applicable. Thanks!

pwadsworth:
The practice exercises documentation says hints.md, and introduction.md are optional. https://exercism.org/docs/building/tracks/practice-exercises#h-documentation-files

ceddlyburge:
Hi Pete. Thanks for looking in to hints.md.

The fact that the instructions are difficult to understand is extremely useful feedback, so thank you for that. And the fact that you are fixing the problem is even better.

Updating things in the problem specifications repo is a bit more annoying, but the plus side is that any improvements will be taken up by all tracks, which is a big deal.

If you create a PR there please let me know, and I'll give it a review.

Thanks! Cedd

PR #642, opened 2024-01-06
Add reverse-string exercise

### LINKED ISSUE #641: Port `reverse-string`?
[Exercism is featuring `reverse-string` starting January 23.](https://forum.exercism.org/t/48in24-exercise-01-23-reverse-string/8978). I'd like to port it before then.

BNAndras:
Closes #641

### DIFF
--- a/config.json
+++ b/config.json
@@ -481,7 +481,7 @@
         "prerequisites": [
           "strings"
         ],
-        "difficulty": 2
+        "difficulty": 1
       },
       {
         "slug": "robot-simulator",
--- a/config.json
+++ b/config.json
@@ -472,6 +472,18 @@
         ]
       },
       {
+        "slug": "reverse-string",
+        "name": "Reverse String",
+        "uuid": "33798c5a-f2ca-4e7b-9dcc-96a24ccf9f24",
+        "practices": [
+          "strings"
+        ],
+        "prerequisites": [
+          "strings"
+        ],
+        "difficulty": 2
+      },
+      {
         "slug": "robot-simulator",
         "name": "Robot Simulator",
         "uuid": "6b8f952e-94ea-4264-accf-cdb28b3b4529",
--- a/exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
+++ b/exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
@@ -3,4 +3,4 @@ module ReverseString exposing (reverse)
 
 reverse : String -> String
 reverse str =
-    String.foldl String.cons "" str
+    String.reverse str
--- a/exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
+++ b/exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
@@ -1,0 +1,6 @@
+module ReverseString exposing (reverse)
+
+
+reverse : String -> String
+reverse str =
+    String.foldl String.cons "" str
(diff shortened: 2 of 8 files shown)


ceddlyburge (approved):
Looks good to me! I'll leave others a bit of time in case they want to review, but if not I'll merge it next weekend (which I think is a week before it is needed for the 48 in 24 challenge)

exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
```diff
@@ -0,0 +1,6 @@
+module ReverseString exposing (reverse)
+
+
+reverse : String -> String
+reverse str =
+    String.foldl String.cons "" str
```
jiegillet, line 6: Why not `String.reverse`?

config.json
```diff
@@ -471,6 +471,18 @@
...
+        "slug": "reverse-string",
+        "name": "Reverse String",
+        "uuid": "33798c5a-f2ca-4e7b-9dcc-96a24ccf9f24",
+        "practices": [
+          "strings"
+        ],
+        "prerequisites": [
+          "strings"
+        ],
+        "difficulty": 2
```
jiegillet, line 484: I think that's more like a 1, for a function that's in the standard lib.

jiegillet (commented):
Thanks for the contribution, looks good.

Just a note, I'm not a fan of this exercise, since it's too basic (standard lib function). In the Elixir track, we decided not to implement it for the same reason. I don't mind merging too much though, since the 48in24 people will expect it 🤷

ceddlyburge:
I do agree that this is a pointless exercise, and it seems like a strange choice for the 48 in 24. Maybe some languages don't have an equivalent of String.reverse, in which case we could ask students to solve it without using the library function?

BNAndras:
I could add an instructions append to recommend not using the standard library. Strain has similar text in the instructions that we can adapt. I avoided using the built-in reverse to show the exercise is fairly straightforward even without it.

ceddlyburge:
Adding something to the instructions sounds like a good idea to me.

We could even enforce it with the analyzer. Not sure it is worth it though.

BNAndras:
I agree. I'm skeptical that anything to curtail standard library use beyond the append would justify the effort spent. If an append won't suffice, I'm fine not porting the exercise.

ceddlyburge:
Adding something to the instructions works for me ...

jiegillet:
I don't think this requires adding special instructions, it's not worth it. `List.reverse` is how you'd solve that problem, so be it. Quick and easy learning exercise.
If you want to implement interesting list-style operations, we have `list-ops`.

exercises/practice/reverse-string/.meta/src/ReverseString.example.elm
BNAndras: I can switch it to String.reverse, but I wanted to make sure it was solvable other ways. This is a practice exercise so my view is that the example is primarily there for CI purposes. For a concept exercise, I'd use String.reverse since mentors would see the code.

That said, I can change it since I've proven at least to myself there are other solutions to the problem.

config.json
BNAndras: Can do. Typically, I avoid 1 on my own tracks since that's the `hello-world` difficulty, and all exercises are at least slightly more difficult than that. :)

jiegillet: Haha that's a valid point, oh well, this expresses well my view on this exercise :)

jiegillet (approved):

PR #665, opened 2024-03-09
Add new practice exercise `sieve`

jiegillet:
It's been a while since I've implemented an exercise :)
This one will be featured on week 11 of 48in24.

This exercise is deceptively complex, because you are supposed to find prime numbers without the use of the division or modulo operator. It's very easy to solve otherwise, but then it wouldn't be the Sieve of Eratosthenes, so it will require an Elixir Analyzer rule, which I will work on next (EDIT: [here](https://github.com/exercism/elm-analyzer/pull/86)).


bin/generate_practice_exercise.sh
```diff
@@ -20,7 +20,7 @@ echo "Fetching latest version of configlet..."
 echo "Adding instructions and configuration files..."
 UUID=$(bin/configlet uuid)
 jq --arg slug "$SLUG" --arg uuid "$UUID" \
-    '.exercises.practice += [{slug: $slug, name: $slug, uuid: $uuid, practices: [], prerequisites: [], difficulty: 5}]' \
```
ceddlyburge, line 23: Is this to avoid a casing type problem with the title of the exercise?

exercises/practice/sieve/.meta/src/Sieve.example.elm
```diff
@@ -0,0 +1,61 @@
...
+
+runSieve : Maybe Int -> Array Bool -> Array Bool
+runSieve maybePrime sieve =
+    case maybePrime of
+        Nothing ->
+            sieve
+
+        Just prime ->
+            let
+                removeMultiples : Int -> Array Bool -> Array Bool
```
ceddlyburge, line 36:
```suggestion
                removeMultiples : Int -> (Array Bool -> Array Bool)
```

I think the brackets make this clearer

ceddlyburge (approved):
I haven't tried to solve it (and not sure I want to!) but all the changes look good.

bin/generate_practice_exercise.sh
jiegillet: Exactly, it's too easy to miss when the name is prefilled, it happened a lot that people didn't change it (me included). At least with a big TODO you can't miss it (hopefully).

jiegillet:
Let's merge this when the rule is merged (or week 11 is about to start)

PR #666, opened 2024-03-09
Add practice exercise `minesweeper`

jiegillet:
Here's another one, while I was at it...
This is week 18 of 48in24.

Some small points of note:
- I didn't implement one test with a contrived empty board (`"\n"`), because there was already a perfectly good one (`""`)
- The boards look very ugly in the tests, but they are in the representation that makes the most sense to me

### DIFF
--- a/exercises/practice/minesweeper/.docs/instructions.md
+++ b/exercises/practice/minesweeper/.docs/instructions.md
@@ -1,0 +1,32 @@
+# Instructions
+
+Add the mine counts to a completed Minesweeper board.
+
+Minesweeper is a popular game where the user has to find the mines using numeric hints that indicate how many mines are directly adjacent (horizontally, vertically, diagonally) to a square.
+
+In this exercise you have to create some code that counts the number of mines adjacent to a given empty square and replaces that square with the count.
+
+The board is a rectangle composed of blank space (' ') characters.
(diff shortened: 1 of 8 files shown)


exercises/practice/minesweeper/.docs/instructions.md
```diff
@@ -0,0 +1,32 @@
+# Instructions
+
+Add the mine counts to a completed Minesweeper board.
+
+Minesweeper is a popular game where the user has to find the mines using numeric hints that indicate how many mines are directly adjacent (horizontally, vertically, diagonally) to a square.
+
+In this exercise you have to create some code that counts the number of mines adjacent to a given empty square and replaces that square with the count.
```
ceddlyburge, line 7:
```suggestion
In this exercise you have to create some code that counts the number of mines adjacent to a given empty square and replace that square with the count.
```

I realise this probably comes from the problem-specifications, but this looks like a typo to me.

ceddlyburge (approved):
I didn't try to solve the exercise, but it all looks good to me ...

exercises/practice/minesweeper/.docs/instructions.md
jiegillet: it's a little awkward, but I think it's correct:

> you have to create some code the counts [...] and replaces [...]

PR #667, opened 2024-03-10
Fix links in allergies approaches

ceddlyburge:
It would be good to lint the markdown, its very hard to spot mistakes without releasing and checking on the live website. It would also help with the annoying thing where every sentence is supposed to be on its own line.

Anyway, this PR hopefully fixes the problems I found the the approaches for the allergies exercise that was recently merged.


jiegillet (approved):
I agree with you about the markdown linter. Other tracks also have link checkers, although those are usually a bit flaky.

Could you open an issue for this? We can investigate the most practical way to do this.

PR #670, opened 2024-03-17
Give task sections more explicit names

MatthijsBlom:
Intended effect in the output of `elm-test`:

```diff
 > Tests
 > Squeaky Clean
-> 5
+> Task 5
 > Omit digits
 
     "1My2Finder3"
     ╷
     │ Expect.equal
     ╵
     "MyFinder"
```

Beware: if this is merged the test runner might need some adjustment as well.

I used to ignore the digits with unclear origin; the aim of this change is to make their meaning very clear.

Contributor to previous confusion: e.g. `squeaky-clean` has several tests named `"Omit digits"`, in `elm-test`'s output only distinguished by section number.

[no important files changed]

### DIFF
--- a/exercises/concept/annalyns-infiltration/tests/Tests.elm
+++ b/exercises/concept/annalyns-infiltration/tests/Tests.elm
@@ -8,7 +8,7 @@ import Test exposing (..)
 tests : Test
 tests =
     describe "AnnalynsInfiltration"
-        [ describe "1"
+        [ describe "Task 1"
             [ test "Cannot execute fast attack if knight is awake" <|
                 \_ ->
                     let
@@ -26,7 +26,7 @@ tests =
                     canFastAttack knightIsAwake
                         |> Expect.equal True
             ]
-        , describe "2"
+        , describe "Task 2"
             [ test "Cannot spy if everyone is sleeping" <|
                 \_ ->
                     let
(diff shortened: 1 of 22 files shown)


mpizenberg:
If going in this direction, some changes would need to happen in the test runner of elm-test-rs indeed. For information, the parsing of the description hierarchy and encoding of exercism-specific format happens in this module https://github.com/mpizenberg/elm-test-runner/blob/19ca6e228828e15714f8a307fae87714a4bfe6c3/src/ElmTestRunner/Reporter/Exercism.elm#L152

ceddlyburge:
Maybe there is a way to just add "Task " in front of the task id in the test runner?

MatthijsBlom:
That doesn't help those who run the tests locally.

The website interface is plenty clearer already: [Squeaky Clean editor](https://exercism.org/tracks/elm/exercises/squeaky-clean/edit).

jiegillet:
I'm going to close this PR since we determined that it requires change in other repositories, and it's not clear that it is going to happen. We can open it again if it does happen.

My personal opinion is that the benefits ("Task 5" instead of "5" for local users) are not really worth the effort since concept exercises are really made to be used in the web editor (putting concept explanations right there in the UI, running the analyzer...).

PR #676, opened 2024-04-03
Use Parser.problem instead of Debug.todo

ceddlyburge:
As per https://forum.exercism.org/t/debug-todo-in-paolas-prestigious-pizza/10596

I ran the existing code locally and it did reproduce the problem reported by Glenn.

I then used Parser.problem instead as suggested, and the test failure messages respond with something useful:

```
Tests
↓ PaolasPrestigiousPizza
↓ 7
✗ parse a pizza with no ingredients

    Err [{ col = 1, problem = Problem "Please implement this parser", row = 1 }]
    ╷
    │ Expect.equal
    ╵
    Ok { ingredients = [], name = "triste", price = 1, vegetarian = True }
```

### DIFF
--- a/exercises/concept/paolas-prestigious-pizza/src/PaolasPrestigiousPizza.elm
+++ b/exercises/concept/paolas-prestigious-pizza/src/PaolasPrestigiousPizza.elm
@@ -22,34 +22,34 @@ type alias Pizza =
 
 priceParser : Parser Int
 priceParser =
-    Debug.todo "Please implement this parser"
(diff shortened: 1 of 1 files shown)


mpizenberg:
Sounds like a nice improvement indeed. Circumventing the usage of `Debug.todo` with function that do not have arguments and making the program crash. Good find!

mpizenberg (approved):

jiegillet (approved):
Beautiful!

Issue #684, opened 2024-04-24
Add test cases to excercise Gotta Snatch'Em All


nico-or:
In part 7, the instructions is "Implement totalCards, which takes a list of collections and returns the total number of different cards in the all of the collections." 

The function signature is:
`totalCards : List (Collection) -> Int`

But reading the test cases, there is no cover for inputs of singleton Lists.

I think adding these inputs should suffice:
- `[Set.empty]`
- `[Set.singleton "Shazam"]`
- `[Set.fromList [ "Shazam", "Wigglycream" ]]`

On the other hand, since this exercise is intended to teach Sets, I don't know if focusing on the proper handling of the List input is appropriate.

jiegillet:
Sure, I think we can add one test in that vein, something like

```elm
            , test "totalCards of a single collection" <|
                \() ->
                    GottaSnatchEmAll.totalCards [ Set.fromList [ "Shazam", "Wigglycream" ]]
                        |> Expect.equal 2
```

Would you like to do it?

PR #686, opened 2024-04-26
Add test case for Gotta Snatch'Em All

nico-or:
Add a test case for totalCards using a singleton list as input since there it wasn't any before.

Discussed in issue #684

### DIFF
--- a/exercises/concept/gotta-snatch-em-all/tests/Tests.elm
+++ b/exercises/concept/gotta-snatch-em-all/tests/Tests.elm
@@ -157,6 +157,10 @@ tests =
                         |> Expect.equal 0
             , test "totalCards of empty collections" <|
                 \() ->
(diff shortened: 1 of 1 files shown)


nico-or:
I wasn't aware of the [no important files changed] flag. Should I add it to the commit body to avoid the re-run of tests?

jiegillet:
> I wasn't aware of the [no important files changed] flag. Should I add it to the commit body to avoid the re-run of tests?

No don't, this new test could potentially reveal issues with some solutions, rerunning the tests is what we want :)

jiegillet:
@nico-or any reason why this is still a draft?

nico-or:
> @nico-or any reason why this is still a draft?

No, I just forgot to switch the status.

jiegillet (approved):
Thank you 👍

PR #691, opened 2024-05-18
Add new practice exercise `pig-latin`

jiegillet:
This one is coming up in week 20 of 48in24.

Not the most interesting exercise, because it's all about exceptions to a weird rule, but it's not the worst either. I didn't set anything in the `"practices"` field because it's clearly all about strings, but we have way too many in that category already.

### DIFF
--- a/exercises/practice/pig-latin/.meta/src/PigLatin.example.elm
+++ b/exercises/practice/pig-latin/.meta/src/PigLatin.example.elm
@@ -1,0 +1,56 @@
+module PigLatin exposing (translate)
+
+
+translate : String -> String
+translate =
+    String.words >> List.map translateWord >> String.join " "
+
+
+translateWord : String -> String
+translateWord input =
+    if String.startsWith "xr" input || String.startsWith "yt" input then
+        input ++ "ay"
+
+    else if String.startsWith "y" input then
+        String.dropLeft 1 input ++ "yay"
+
+    else
+        let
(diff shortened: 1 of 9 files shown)


exercises/practice/pig-latin/.meta/src/PigLatin.example.elm
```diff
@@ -0,0 +1,56 @@
...
+        isVowelOrY char =
+            List.member char [ 'a', 'e', 'i', 'o', 'u', 'y' ]
+
+        ( first, second ) =
+            input |> String.toList |> takeWhile (isVowelOrY >> not)
+    in
+    ( String.fromList first, String.fromList second )
+
+
+takeWhile : (a -> Bool) -> List a -> ( List a, List a )
```
ceddlyburge, line 41: This is picky, but I would probably rename this function. takeWhile doesn't seem to be what it is doing really. Maybe `split` or `slice` or something like that?

ceddlyburge (approved):

exercises/practice/pig-latin/.meta/src/PigLatin.example.elm
jiegillet: It takes the first part of a list while a condition is true. The only difference with the usual `takeWhile` is that it returns both parts of the list, and that's clear with the type signature. I think the name makes sense.

PR #696, opened 2024-05-31
Add practice exercise `spiral-matrix`

jiegillet:
Week 26 of 48in24

Not the most interesting :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1280,6 +1280,22 @@
           "booleans"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "spiral-matrix",
+        "name": "Spiral Matrix",
+        "uuid": "e44952ec-185a-4c8e-9404-fb7b365e3961",
+        "practices": [],
+        "prerequisites": [
+          "lists",
+          "comparison",
+          "dict",
+          "let",
+          "pattern-matching",
+          "tuples",
+          "custom-types"
+        ],
+        "difficulty": 6
       }
     ]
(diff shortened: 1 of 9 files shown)


config.json
```diff
@@ -1280,6 +1280,22 @@
           "booleans"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "spiral-matrix",
+        "name": "Spiral Matrix",
+        "uuid": "e44952ec-185a-4c8e-9404-fb7b365e3961",
+        "practices": [],
```
jiegillet, line 1288: Note for later: could be a good candidate for  #692

exercises/practice/spiral-matrix/src/SpiralMatrix.elm
```diff
@@ -0,0 +1,6 @@
+module SpiralMatrix exposing (spiralMatrix)
+
```
ceddlyburge, line 2: Should we provide the 'Direction' type here? 

It's just a thought ...

ceddlyburge (approved):

exercises/practice/spiral-matrix/src/SpiralMatrix.elm
jiegillet: No, I don't think so, that was just my own way of solving it, there are infinitely many other ways to do it.
Also, it's not in the instructions so it would be confusing.

PR #700, opened 2024-06-14
Add Knapsack

kahgoh:
This is one of the [48in24 exercises](https://exercism.org/challenges/48in24/implementation_status).

### DIFF
--- a/config.json
+++ b/config.json
@@ -1321,7 +1321,7 @@
           "lists",
           "records"
         ],
-        "difficulty": 6
+        "difficulty": 7
       }
     ]
   },
--- a/config.json
+++ b/config.json
@@ -1310,6 +1310,18 @@
           "custom-types"
         ],
         "difficulty": 6
+      },
+      {
+        "slug": "knapsack",
+        "name": "Knapsack",
+        "uuid": "9bf186d3-e904-42f2-a705-dffbe98adb08",
+        "practices": [],
+        "prerequisites": [
+          "comparison",
+          "lists",
+          "records"
+        ],
+        "difficulty": 6
       }
     ]
   },
--- a/exercises/practice/knapsack/.meta/src/Knapsack.example.elm
+++ b/exercises/practice/knapsack/.meta/src/Knapsack.example.elm
@@ -1,0 +1,71 @@
+module Knapsack exposing (Item, maximumValue)
+
+
+type alias Item =
+    { value : Int
+    , weight : Int
+    }
+
+
+maximumValue : Int -> List Item -> Int
+maximumValue capacity items =
+    doMaximumValue items (List.repeat (capacity + 1) 0)
+
+
+last : List Int -> Int
+last list =
+    List.foldl (\x _ -> x) 0 list
+
+
+valueAt : List a -> Int -> Result String a
+valueAt list index =
+    case list of
+        [] ->
+            Err "Index out of bounds"
+
+        n :: rem ->
+            if index == 0 then
+                Ok n
+
+            else
+                valueAt rem (index - 1)
+
+
+nextRow : Item -> List Int -> List Int
+nextRow item lastRow =
+    List.indexedMap
+        (\cap val ->
+            let
+                req =
+                    cap - item.weight
+            in
+            if req < 0 then
+                val
+
+            else
+                case valueAt lastRow req of
+                    Ok v ->
+                        let
+                            withItem =
+                                v + item.value
+                        in
+                        if withItem > val then
+                            withItem
+
+                        else
+                            val
+
+                    Err _ ->
+                        val
+        )
+        lastRow
+
+
+doMaximumValue : List Item -> List Int -> Int
+doMaximumValue items lastRow =
+    case items of
+        [] ->
+            last lastRow
+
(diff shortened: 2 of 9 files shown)


exercises/practice/knapsack/src/Knapsack.elm
```diff
@@ -0,0 +1,12 @@
...
+
+type alias Item =
+    { value : Int
+    , weight : Int
+    }
+
+
+maximumValue : Int -> List Item -> Int
+maximumValue capacity items =
+    Debug.todo "Please implement this function"
```
jiegillet, line 12:
```suggestion
    Debug.todo "Please implement maximumValue"
```

exercises/practice/knapsack/tests/Tests.elm
```diff
@@ -0,0 +1,85 @@
...
+import Knapsack exposing (..)
+import Test exposing (..)
+
+
+tests : Test
+tests =
+    describe "Knapsack"
+        [ test "no items" <|
+            \() ->
+                Expect.equal 0 <| maximumValue 100 []
```
jiegillet, line 13: We usually format our tests in this way
```suggestion
            maximumValue 100 []
            |> Expect.equal 0
```
It feels more like English.
(this applies to all tests)

config.json
```diff
@@ -1310,6 +1310,18 @@
           "custom-types"
         ],
         "difficulty": 6
+      },
+      {
+        "slug": "knapsack",
+        "name": "Knapsack",
+        "uuid": "9bf186d3-e904-42f2-a705-dffbe98adb08",
+        "practices": [],
```
jiegillet, line 1318: [You don't need to do anything here] This is also a good candidate for #692

exercises/practice/knapsack/.meta/src/Knapsack.example.elm
```diff
@@ -0,0 +1,71 @@
...
+maximumValue capacity items =
+    doMaximumValue items (List.repeat (capacity + 1) 0)
+
+
+last : List Int -> Int
+last list =
+    List.foldl (\x _ -> x) 0 list
+
+
+valueAt : List a -> Int -> Result String a
```
jiegillet, line 20: This could be a `Maybe a`, since you don't really do anything with the error

exercises/practice/knapsack/.meta/src/Knapsack.example.elm
```diff
@@ -0,0 +1,71 @@
...
+valueAt list index =
+    case list of
+        [] ->
+            Err "Index out of bounds"
+
+        n :: rem ->
+            if index == 0 then
```
jiegillet, line 27: This doesn't really matter, but in this situation I would generally do
```suggestion
            if index <= 0 then
```
to avoid infinite recursion if someone enters a negative index (which I know won't happen here)

exercises/practice/knapsack/.meta/src/Knapsack.example.elm
```diff
@@ -0,0 +1,71 @@
...
+            else
+                valueAt rem (index - 1)
+
+
+nextRow : Item -> List Int -> List Int
+nextRow item lastRow =
+    List.indexedMap
+        (\cap val ->
+            let
+                req =
```
jiegillet, line 39: Three letter variable names like `cap`, `v`, `value`, `req` are hard to read.

exercises/practice/knapsack/.meta/src/Knapsack.example.elm
```diff
@@ -0,0 +1,71 @@
...
+                    Ok v ->
+                        let
+                            withItem =
+                                v + item.value
+                        in
+                        if withItem > val then
+                            withItem
+
+                        else
+                            val
```
jiegillet, line 56: Maybe less readable, but I think this whole branch could be `max val (v + item.value)`

config.json
```diff
@@ -1310,6 +1310,18 @@
...
+        "prerequisites": [
+          "comparison",
+          "lists",
+          "records"
+        ],
+        "difficulty": 6
```
jiegillet, line 1324: Feel like a 7 or an 8, because the algorithm isn't trivial. What do you think?

jiegillet (commented):
Fantastic, thank you for the contribution!
I left a bunch of nitpicks because I'm a grumpy nitpicky man, but I'll be happy to merge this :)

config.json
kahgoh: I think a 7 is also reasonable. My hesitance with going to an 8 is seeing Zebra puzzle at 8 (and me thinking its not quite as hard Zebra Puzzle). The other thought I have is this example solution uses memoization. A simpler solution could be to use recursion without the memoization (although it won't be as efficient).

kahgoh:
Thanks for the review @jiegillet! I was happy with the suggestions and have pushed the updates.

jiegillet (approved):
Very nice, thanks again.

config.json
jiegillet: Yes you're right, it's not quite a Zebra puzzle.
In the end it's not an exact science, but 7 feels right :)

PR #705, opened 2024-07-21
Add new practice exercise `meetup`

jiegillet:
This is week 32 of 48in24.

This one was fun, mostly because Elm's `Time` module is completely useless here, so you'll need to research algorithms and actually think for a bit.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1296,6 +1296,21 @@
         "difficulty": 4
       },
       {
+        "slug": "meetup",
+        "name": "Meetup",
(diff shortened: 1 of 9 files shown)


ceddlyburge (approved):
I didn't take the time to fully understand this, but it looks it does what it should. Send like it practices weird date stuff not then elm though :)

jiegillet:
Thank you!
Yeah, this exercise would a lot faster with access to a date library, but I did find it quite fun to do anyway :)

Issue #639, opened 2023-12-07
Concept - Web applications part 1


ceddlyburge:
# Design

## Goal

The Goal is to learn how to write basic web applications in Elm.

## Learning objectives

Students should be able to
- understand The [Elm Architecture pattern](https://guide.elm-lang.org/architecture/)
- define an appropriate Model type
- define an appropriate Msg type 
- create a Browser.sandbox application
- write an update function
- write a view function

## Out of scope

- Http calls
- Json encoding / decoding
- Javascript interop
- Commands and subscriptions

## Concepts

The concept this exercise unlocks is:

- `web-applications-1`

## Prerequisites

- records
- custom-types
- pattern-matching
- lists
- maybe partial-application (The [Elm Guide forms section](https://guide.elm-lang.org/architecture/forms) uses this)

## Analyzer

TBD depending on the exercise

PR #723, opened 2024-11-11
DRAFT: Web Applications part 1 concept

ceddlyburge:
Hi everyone, I'd like to get some early feedback on this.

Its a plan to create a concept for #639.

It is possible to shim the Browser module and the tests all run and pass and test all the important things.

There is a change to the elm.json, but just moving the Html dependency from indirect to direct, so I don't think we would need to update the docker images or anything (although we should update all the other elm.json's for consistency).

I think that any syllabus for learning Elm needs to include actually writing a web app, so I am keen to do this, but want to hear peoples thoughts.


mpizenberg:
> It is possible to shim the Browser module and the tests all run and pass and test all the important things.

Hum, I’ve never used it, but I wonder what you can get with elm-program-test https://github.com/avh4/elm-program-test
Or something like it.

ceddlyburge:
That is an option as well. It would require us to update the docker image for the test runner, but we can definitely do that.

I don't think we need it for this exercise though, we can test virtually everything already.

THe only thing we can't test is that the correct update function gets passed to Browser.sandbox, but we do know that it has the correct type signature, and we do test the update function itself, so you would have to write some very weird code for it to be a problem.

jiegillet:
This is nice, and you are testing all that needs to be tested, although I would also give elm-program-test a shot, maybe we can get even better tests (and they would also be more realistic).

However, I'm still unsure whether that concept really works. The point of the Browser module is to make web pages, but here you will never see it, that seems like a shame. Maybe it's still good to learn about the Elm architecture in theory, but... yeah, I'm not sure.

ceddlyburge:
I've never used elm-program-test, and Matthieu hasn't either, not sure if you have?

But I do test my update and view functions routinely, and sometimes my init functions if they are complicated. So I think the testing that is done here is basically the same as what I normally do, so it feels quite realistic to me.

It is true that you would never see the code running in the online editor, but we could promote Ellie, and people could copy and paste it in to there to see it running. We could also give people instructions to make it work locally on peoples computer (delete Browser.elm and install the Browser package).

Another option is to use the actual Browser module, instead of using a fake. This would mean that if you were solving the exercise on your own computer, you could run it and see a web page. It feels like it might open up a security hole though, but I could investigate if people think this is worth pursuing.

ceddlyburge:
We could maybe try and have different files (Browser.elm) for the online version and the download version (no Browser.elm).

I imagine this would require a chanage to exercism, but I could put something on the forum

ceddlyburge:
Ok cool, I'll try and keep the html and css part as small as possible, and keep going with this. We can always see how we like it later, or see if we get any feedback about it if it goes live

mpizenberg:
Nice! I agree with minimal html/css stuff and adjust after feedback. Maybe it will also be pioneering some new capabilities for exercism on visualization for web languages tracks.

PR #724, opened 2024-12-09
Sync docs, metadata and tests

jiegillet:
This PR syncs the docs, metadata, and adds two new tests (one is a re-implementation).

### DIFF
--- a/exercises/practice/bob/tests/Tests.elm
+++ b/exercises/practice/bob/tests/Tests.elm
@@ -127,8 +127,8 @@ tests =
             test "on multiple line questions" <|
                 \() ->
                     Expect.equal
-                        "Whatever."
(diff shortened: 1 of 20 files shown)


exercises/practice/bob/tests/Tests.elm
```diff
@@ -127,8 +127,8 @@ tests =
             test "on multiple line questions" <|
                 \() ->
                     Expect.equal
-                        "Whatever."
-                        (Bob.hey "\nDoes this cryogenic chamber make me look fat?\nno")
+                        "Sure."
```
ceddlyburge, line 130: This looks like the instructions would have to change too?

jiegillet: Nope, now it ends with a question mark, so the expected answer is different, as per the instructions.
I'm not sure what motivated this particular change, you'd have to check on the problem-specs repo.

PR #725, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

ceddlyburge (approved):
That forum post is very long, I'm just going to assume that the final decision was a good one and approve this.

PR #729, opened 2025-02-08
Fix Random.weighted bug in the concept doc: weight is the first element of the tuple

glennj:
Forum thread: 
http://forum.exercism.org/t/random-concept-random-weighted-tuples-are-backwards/15566

Ref `weighted` doc: https://package.elm-lang.org/packages/elm/random/latest/Random#weighted


ceddlyburge (approved):
Hi Glenn, I have now taken a proper look at this, and you're right, this was the wrong way round.

Issue #630, opened 2023-11-09
Fill all exercises with accurate "practices" and "prerequisites" concepts


jiegillet:
Now that we have more concepts, we should tag exercises with them (like in #629).

The `"practices"` should only have one, maybe two, core concept that will help the students master a specific concept.

The `"prerequisites"` are the concepts necessary to finish the task. This is not an exact science, a rule of thumb is looking at the concepts used in the examples solution.

Using `bin/configlet info` provides valuable information on concepts that are not practiced or prerequisites.
 Ideally, all concepts should be practiced, although it's not always possible (concepts like `phantom-types` don't really make sense for one-module exercises).
We should definitely not have any "Concepts that are a prerequisite, but aren't practiced by any Practice Exercise".

PR #733, opened 2025-03-08
Add new practice exercise `swift-scheduling`

jiegillet:
This is a brand new exercise that looked interesting so I implemented it.
It's entirely based on dates, so since `Time` is so limited, I had to implement a lot of things myself, which was fun. The tests look a little strange, but I though spelling out the year and stuff would be a lot more useful than checking a Posix integer, it certainly helped me when I was solving it.

I am now thinking that we could have a `time` concept, I think I'll open an issue. It would be pretty short.

I also played around with `config.json` (part of #630). Maybe it could have been a separate PR...


config.json
```diff
@@ -387,52 +387,50 @@
         "name": "Bob",
         "uuid": "11ce2a73-3f43-4b14-b755-6c52dc71bdda",
         "practices": [
-          "strings"
+          "booleans"
```
ceddlyburge, line 390: I'm not sure it really makes sense for me to check all these, so maybe just let me know how confident you are.

ceddlyburge (commented):
Hi Jeremie, I'm not sure about the solution to this exercise. What it is mainly doing is showing that it is stupidly hard in Elm to work with dates and times. Which isn't really true, most people just reach for one of the helper packages. I think we shouldn't give students the impression that this is how to work with dates in Elm, because it isn't what it makes sense to do in most situations.
Maybe we could add one of the common date package things instead to make the solution more ergnomic / idiomatic?
I think this is the one I normally use
https://package.elm-lang.org/packages/justinmimbs/date/latest/

exercises/practice/swift-scheduling/.meta/src/SwiftScheduling.example.elm
```diff
@@ -0,0 +1,361 @@
...
+            Now
+
+
+
+-- DATETIME
+
+
+type alias DateTime =
+    { year : Int
+    , month : Month
```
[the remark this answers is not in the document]
jiegillet: Going though `Month` is unavoidable, since that's what `Time.toMonth` gives you, but you are right that an `Int` would have been a good choice too. I like the type safety so I want with this, but ultimately since this is only an example, that's entirely the student's choice.

jiegillet:
> Hi Jeremie, I'm not sure about the solution to this exercise. What it is mainly doing is showing that it is stupidly hard in Elm to work with dates and times. Which isn't really true, most people just reach for one of the helper packages. I think we shouldn't give students the impression that this is how to work with dates in Elm, because it isn't what it makes sense to do in most situations. Maybe we could add one of the common date package things instead to make the solution more ergnomic / idiomatic? I think this is the one I normally use https://package.elm-lang.org/packages/justinmimbs/date/latest/

You say stupidly hard, I say stupidly fun :)
I had never implemented a `DateTime -> Posix` function before, I loved it.

That being said, you might be right. I don't think we should import a new package for this, so maybe we should simply chose not to implement this exercise? Would that be your recommendation?
And what's your opinion on a `time` concept in this context? Not worth doing?

ceddlyburge:
I think it would be less fun if you were panicking to meet a deadline :)

It's a good point though, some people will be on exercism to try interesting challenges, whereas others will be trying to learn a new language as quickly as possible, so maybe we should cater to both.

So maybe we keep the exercise but make it very obvious that its just for fun, and wouldn't be a real world thing, and that you would use a library instead?

I think a Time concept is a good idea. Elm is very different to most other languages with this. But I think for the concept you would definitely want to show students the ergonomic / idiomatic way of doing things, which I think requires us to add whichever package we think is best. I'm open to suggestions though.

jiegillet:
> I think it would be less fun if you were panicking to meet a deadline :)

Let's not bring real world stress to Exercism :)

> It's a good point though, some people will be on exercism to try interesting challenges, whereas others will be trying to learn a new language as quickly as possible, so maybe we should cater to both.
> 
> So maybe we keep the exercise but make it very obvious that its just for fun, and wouldn't be a real world thing, and that you would use a library instead?

Wise advice, I added an instruction append, let me know if that works.

> I think a Time concept is a good idea. Elm is very different to most other languages with this. But I think for the concept you would definitely want to show students the ergonomic / idiomatic way of doing things, which I think requires us to add whichever package we think is best. I'm open to suggestions though.

I'm not sure we should use elm community packages for Exercism. The primary goal for Exercism is to teach languages, things like syntax and concepts. Once you've learned that, you should be able to use any package that's useful for your applications.If we start exploring packages, our task will never be done.
That being said, you're definitely not wrong in wanting to present an ergonomic way of using dates and time. `elm/time` is very limited since it's only meant to display time, not manipulate it. Maybe if that's how I would approach it, by stating that it's only meant to display time provided by the browser.

exercises/practice/swift-scheduling/.docs/instructions.append.md
```diff
@@ -0,0 +1,6 @@
+# Instructions Append
+
+```exercism/note
+`elm/time` is not the most complete package when it comes to handling date and time operations.
```
jiegillet, line 4: @ceddlyburge  like this?
```suggestion
`elm/time` is meant to display a time provided by the browser, not to manipulate dates and times.
```

exercises/practice/swift-scheduling/.docs/instructions.append.md
```diff
@@ -0,0 +1,6 @@
...
+In real world applications, one would reach out to packages from the Elm community, but since those are not accessible in the online editor, take this exercise as a challenge rather than as a practice of idiomatic Elm.
```
ceddlyburge, line 5:
```suggestion
In real world applications, one would reach out to packages from the Elm community to manipulate dates and times, but since those are not accessible in the online editor, take this exercise as a challenge rather than as a practice of idiomatic Elm.
```

PR #734, opened 2025-03-18
Update maze maker introduction

ceddlyburge:
Another pull request updated about.md, but we forgot about this file

Related to #729


jiegillet (approved):
Nice

PR #738, opened 2025-04-20
Update instructions.md

Tagir-A:
Since Instructions explicitly mention "You have three tasks", it seems logical that the text should match the numbered list.


ceddlyburge (approved):
Makes sense, thanks for pointing it out :)

Issue #739, opened 2025-04-21
Add additional practice exercises?


BNAndras:
I'd like to potentially add the following exercises over the next few weeks.

- [x] [darts](https://exercism.org/exercises/darts)
- [x] [relative-distance](https://exercism.org/exercises/relative-distance)
- [x] [resistor-color](https://exercism.org/exercises/resistor-color)
- [x] [resistor-color-duo](https://exercism.org/exercises/resistor-color-duo)
- [x] [resistor-color-trio](https://exercism.org/exercises/resistor-color-trio)
- [x] [square-root](https://exercism.org/exercises/square-root)

jiegillet:
Use the generator and go for it, I'll review them :)

BNAndras:
How about I implement [Relative Distance](https://exercism.org/exercises/relative-distance)? I've done that for 7 tracks to date with another one in progress.

jiegillet:
Sure! I implemented the Elixir version, but I didn't do Elm.

PR #740, opened 2025-04-29
Add `resistor-color`

BNAndras:
Related to #739


exercises/practice/resistor-color/.meta/src/ResistorColor.example.elm
```diff
@@ -0,0 +1,25 @@
...
+    findIndex color colors 0
+
+
+colors : List String
+colors =
+    [ "black", "brown", "red", "orange", "yellow", "green", "blue", "violet", "grey", "white" ]
+
+
+findIndex : a -> List a -> Int -> Int
+findIndex needle haystack acc =
```
ceddlyburge, line 15: I would probably use filter for this, but I don't think it matters either way.

PR #741, opened 2025-04-29
add `resistor-color-duo`

BNAndras:


BNAndras:
Related to #739

BNAndras:
Aha! Third time was the charm with the CI. 🚀

exercises/practice/resistor-color-duo/src/ResistorColorDuo.elm
```diff
@@ -0,0 +1,6 @@
+module ResistorColorDuo exposing (value)
+
+
+value : List String -> Int
```
jiegillet, line 4: Why not use types here? Wouldn't that be more idiomatic? Colors don't need to be string.
I just looked at #740, and I think I would have had the same comment. Maybe we could change it too since it's was just released? @ceddlyburge What do you think?

BNAndras: That seems reasonable to me. I’ll be traveling tonight but I can make both edits tomorrow. I doubt there’ll be a lot of solutions broken this earlier.

ceddlyburge: Hi both, I think it depends how you think about it, and I'm not sure either way is definitely better / worse than the other. At the boundary of the system everything is untyped, so if you think about the exercise like that, then strings make sense. If you think about it being an internal part of a system then types are more idiomatic. But its only an exercise, so it is neither of these situations really. So I think the question is probably which style makes for the better exercise. I don't have a strong opinion on this, I think both ways are already practiced in other exercises, and both have their interest.

jiegillet: My inclination here is towards using types, because that's one of the ingredients that make Elm special, and I would like to showcase it to new users as soon as possible. And since this series is pretty basic, they show up in the curriculum early.

In terms of the algorithm, it really makes no difference for any of the `resistor-color` trio... @BNAndras, let's say it's your call :)

BNAndras: I’m for it. I did something similar for Space Age on the Arturo track where the exercise uses the built-in support for physical quantities to model the ages on each planet. It didn’t make a big difference, but it eased people into an interesting corner of the language.

exercises/practice/resistor-color-duo/src/ResistorColorDuo.elm
```diff
@@ -0,0 +1,6 @@
+module ResistorColorDuo exposing (Color(..), value)
```
jiegillet, line 1: Same as in the other PR, I think we should provide the type definition in the stub. It's just busy work to add them anyway.

exercises/practice/resistor-color-duo/.meta/src/ResistorColorDuo.example.elm
```diff
@@ -0,0 +1,58 @@
...
+
+
+value : List Color -> Int
+value colors =
+    case colors of
+        first :: second :: _ ->
+            colorCode first * 10 + colorCode second
+
+        _ ->
+            -1
```
jiegillet, line 24: This is not in the specs, so I don't know if it's the right approach or not. I guess if it's not tested, it doesn't matter 🤷

config.json
```diff
@@ -450,6 +450,19 @@
         ],
         "difficulty": 2
       },
+      {
+        "slug": "resistor-color-duo",
+        "name": "Resistor Color Duo",
+        "uuid": "a6b5f8e0-96a5-4da1-97ad-033a2a4b115b",
+        "practices": [
+          "lists"
```
jiegillet, line 458: Let's add pattern matching.
Let's add pattern matching and custom types to prerequisites as well.

jiegillet (commented):
Nice, custom types make me happy 😀

exercises/practice/resistor-color-duo/.meta/src/ResistorColorDuo.example.elm
BNAndras: We could use Maybe here instead.

BNAndras: That'd naturally also apply to `resistor-color` and `resistor-color-trio`.

jiegillet: I don't think it's worth it.

jiegillet (approved):
This is great, thank you :)

PR #742, opened 2025-04-29
add `darts`

BNAndras:


BNAndras:
Also related to #739

exercises/practice/darts/tests/Tests.elm
```diff
@@ -0,0 +1,79 @@
+module Tests exposing (tests)
+
+import Darts
+import Expect
+import Test exposing (Test, describe, skip, test)
+
+
+
+-- Return the correct amount earned by a dart's landing position.
```
jiegillet, line 9: We can remove the comments, the generator leaves them because it's sometimes useful to see when you implement the tests.

exercises/practice/darts/.meta/src/Darts.example.elm
```diff
@@ -0,0 +1,20 @@
+module Darts exposing (score)
+
+
+score : Float -> Float -> Int
+score x y =
+    let
+        distance =
+            sqrt (x * x + y * y)
```
jiegillet, line 8: This is a micro-optimization, but since `sqrt` is fairly expensive to compute, it's more efficient to calculate the distance squared, and compare it to values like `10 * 10`. Not that it matters much here, feel free to ignore :)

config.json
```diff
@@ -424,6 +424,19 @@
...
+      {
+        "slug": "darts",
+        "name": "Darts",
+        "uuid": "2df81135-6cd2-4420-9116-ff3b8922d09b",
+        "practices": [
+          "comparison",
+          "let"
+        ],
+        "prerequisites": [
+          "comparison"
```
jiegillet, line 436: Since it practices `let`, it should also be a prerequisite.

jiegillet (approved):
Very nice, thank you. I left minor comments, but I'm approving already

exercises/practice/darts/.meta/src/Darts.example.elm
BNAndras: Yeah, I normally do the square root since then the distances for each score band match the instructions, but I'll switch. Not a big deal.

PR #743, opened 2025-04-29
Add `square-root`

BNAndras:
Related to #739

### DIFF
--- a/config.json
+++ b/config.json
@@ -626,6 +626,14 @@
         "difficulty": 3
       },
       {
+        "slug": "square-root",
+        "name": "Square Root",
+        "uuid": "8960514f-17af-4009-b334-d6bfc4722c80",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
+      },
+      {
         "slug": "twelve-days",
         "name": "Twelve Days",
         "uuid": "50770287-c734-46ba-b3a1-c36d31b6cc3c",
--- a/config.json
+++ b/config.json
@@ -629,8 +629,13 @@
         "slug": "square-root",
         "name": "Square Root",
         "uuid": "8960514f-17af-4009-b334-d6bfc4722c80",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "pattern-matching"
+        ],
+        "prerequisites": [
+          "basics-2",
+          "pattern-matching"
+        ],
         "difficulty": 2
       },
       {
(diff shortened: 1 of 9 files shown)


exercises/practice/square-root/.meta/src/SquareRoot.example.elm
```diff
@@ -0,0 +1,27 @@
...
+                            (guess + toFloat radicand / guess) / 2
+                    in
+                    if abs (guess - nextGuess) < 0.0001 then
+                        nextGuess
+
+                    else
+                        improveGuess nextGuess
+            in
+            toFloat radicand
+                / 2
```
jiegillet, line 25: For readability, I think this needs some parenthesis or an intermediate variable name.

exercises/practice/square-root/.meta/src/SquareRoot.example.elm
```diff
@@ -0,0 +1,27 @@
...
+                |> improveGuess
+                |> floor
```
jiegillet, line 27: Wouldn't round be more appropriate here? What if the guess is `4.999` or something? I can't tell if the algorithm guarantees a to remain above the root (but even if it does, with floats you never know for sure).

I've added this exercise in Elixir before and I remember being frustrated at the small number of tests, when it would be so trivial to add a test that checks the first 500 square numbers or whatever. (this is just me venting, we should not add non-canonical tests, although maybe it's worth opening a thread on the forum).

config.json
```diff
@@ -625,6 +625,14 @@
         ],
         "difficulty": 3
       },
+      {
+        "slug": "square-root",
+        "name": "Square Root",
+        "uuid": "8960514f-17af-4009-b334-d6bfc4722c80",
+        "practices": [],
```
jiegillet, line 632: This would be a perfect fit for #692 but in the mean time, we can leave it empty

config.json
```diff
@@ -625,6 +625,14 @@
...
+        "prerequisites": [],
```
jiegillet, line 633: Haha, we really don't have much that could fit here.
Your solution uses pattern matching, so we could add that, also `basics-2` I suppose.

jiegillet (commented):
Thank you and great job, I have some minor comments.

exercises/practice/square-root/.meta/src/SquareRoot.example.elm
BNAndras: That's a good question. This is an adaptation of my example solution for Dart where I didn't really think about edge cases. I'll noodle on this and get back to you.

But yeah it'd be nice to have more test cases. I'm guessing if we identify a test case that would fail for this approach, the other maintainers may be amenable to adding it.

jiegillet (approved):
Perfect, let's ship it 🚀 
Thank you again!

PR #744, opened 2025-05-01
Change `resistor-color` inputs to custom type

BNAndras:
Related to the PR review for #741

### DIFF
--- a/exercises/practice/resistor-color/src/ResistorColor.elm
+++ b/exercises/practice/resistor-color/src/ResistorColor.elm
@@ -1,11 +1,11 @@
-module ResistorColor exposing (colorCode, colors)
+module ResistorColor exposing (Color(..), colorCode, colors)
 
 
-colorCode : String -> Int
+colorCode : Color -> Int
 colorCode color =
     Debug.todo "Please implement colorCode"
 
 
-colors : List String
+colors : List Color
 colors =
     Debug.todo "Please implement colors"
--- a/exercises/practice/resistor-color/src/ResistorColor.elm
+++ b/exercises/practice/resistor-color/src/ResistorColor.elm
@@ -1,6 +1,19 @@
 module ResistorColor exposing (Color(..), colorCode, colors)
 
 
+type Color
+    = Black
+    | Brown
+    | Red
+    | Orange
+    | Yellow
+    | Green
+    | Blue
+    | Violet
+    | Grey
+    | White
+
+
 colorCode : Color -> Int
 colorCode color =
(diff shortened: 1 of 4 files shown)


exercises/practice/resistor-color/src/ResistorColor.elm
```diff
@@ -1,11 +1,11 @@
-module ResistorColor exposing (colorCode, colors)
+module ResistorColor exposing (Color(..), colorCode, colors)
```
jiegillet, line 1: I think we should provide the `Color` type, otherwise the stub will not compile. WDYT?

jiegillet (commented):
Excellent!
Can you add "pattern-matching" and "custom-types" to the prerequisites?
Also I think this exercise now practices pattern matching and lists.

exercises/practice/resistor-color/src/ResistorColor.elm
BNAndras: I also went back and forth on whether the `colors` test should be included. It's trivial to paste from the test data, and we're encouraging students to repeat themselves since the custom type enumerates all the colors anyways.

For the type, it'd be useful if we could provide a starting point like `type Color = ...` that could compile but then warn it's not implemented. Without that, I'd prefer a commented-out line explicitly telling the students to add the Color type. That alerts the student to why the code won't compile and suggests the fix (adding the type).

jiegillet: We sometimes use `type Color = TODO`, but I don't think we can provide a starting point that would make the stub compile, since there is a test that has all the colors variants in it.
I try to avoid comments, because some people don't remove them, and that drives me nuts :p

I'd say that if the task of adding the types is trivial/just busy work, then we might as well do it for them, the point of the exercise is not to model the colors, it's to implement the functions. Also if you're afraid of boilerplate, Elm is not for you :)

But I see your point, so I'll leave it up to you.

Anecdotally, I just noticed that this exercise stub already doesn't compile, I think because `Debug.todo` used for a constant (like for `colors`) is created at compile time, unlike functions. I don't remember the type of trick we use for this though.

BNAndras: It seems other exercises do `Debug.todo` as well.

jiegillet: For constants? Then it's a failure on our part ^^

jiegillet (approved):
Perfect, thank you!

PR #747, opened 2025-05-02
Add `resistor-color-trio`

BNAndras:
Related to #739

### DIFF
--- a/config.json
+++ b/config.json
@@ -470,6 +470,22 @@
         "difficulty": 2
       },
       {
+        "slug": "resistor-color-trio",
+        "name": "Resistor Color Trio",
+        "uuid": "761f7552-c192-4d46-a21c-fa30c4788307",
+        "practices": [
+          "lists",
+          "pattern-matching"
+        ],
+        "prerequisites": [
+          "lists",
+          "basics-2",
+          "custom-types",
+          "pattern-matching"
+        ],
+        "difficulty": 2
+      },
+      {
         "slug": "space-age",
         "name": "Space Age",
         "uuid": "2ce5bcba-c708-4da0-ae11-4654b3e0d0a5",
--- a/config.json
+++ b/config.json
@@ -483,7 +483,7 @@
           "custom-types",
           "pattern-matching"
         ],
-        "difficulty": 2
+        "difficulty": 3
       },
       {
         "slug": "space-age",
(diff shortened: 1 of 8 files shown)


BNAndras:
I'm thinking it'd be useful to add tests for decimal resistor values since 3.3k ohm resistors are a thing. We recently added that on the Emacs Lisp track, and I think those tests were borrowed from Julia.

https://github.com/exercism/emacs-lisp/blob/main/exercises/practice/resistor-color-trio/resistor-color-trio-test.el

exercises/practice/resistor-color-trio/.meta/src/ResistorColorTrio.example.elm
```diff
@@ -0,0 +1,83 @@
...
+    | Violet
+    | Grey
+    | White
+
+
+label : List Color -> String
+label colors =
+    let
+        raw =
+            colors |> rawValue |> toFloat
```
jiegillet, line 21: Why do you need to use floats here? Would `//` not do the job for integer division?

We got a whole thing with floats at work, they make me a bit jumpy now :)

jiegillet (approved):
Looks great, approving already.
I left small non-blocking suggestions.

jiegillet:
> I'm thinking it'd be useful to add tests for decimal resistor values since 3.3k ohm resistors are a thing. We recently added that on the Emacs Lisp track, and I think those tests were borrowed from Julia.

I would rather stick to the canonical data, it really makes life easier in the long term for maintainers.

config.json
```diff
@@ -469,6 +469,22 @@
...
+          "lists",
+          "pattern-matching"
+        ],
+        "prerequisites": [
+          "lists",
+          "basics-2",
+          "custom-types",
+          "pattern-matching"
+        ],
+        "difficulty": 2
```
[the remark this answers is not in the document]
BNAndras: Sure.

exercises/practice/resistor-color-trio/.meta/src/ResistorColorTrio.example.elm
BNAndras: Alright, I wrote out the long-form literals so we don't need floats anymore.

BNAndras:
I renamed a few things and I added some handling for single-band resistor lists just as a nice nod back to `resistor-color` and how we're building up the complexity.

jiegillet:
Awesome :)

PR #749, opened 2025-05-11
Add new practice exercise `split-second-stopwatch`

jiegillet:
This is a brand new one fresh from the problem specs.

Originally, I thought that would be an exercise about time, but really it's much more suited to be our first practice exercise about `opaque-types`.
In reality, `Stopwatch` can be implemented as a `type alias`, but I opted not to do that in the example solution to push students to use proper opaque types. Of course, we can't stop them from using type aliases, but we could add an analyzer check with some appropriate message. What do you think?

### DIFF
--- a/exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
+++ b/exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
@@ -1,0 +1,141 @@
+module SplitSecondStopwatch exposing
+    ( State(..)
+    , Stopwatch
+    , advanceTime
+    , currentLap
+    , lap
+    , new
+    , previousLaps
+    , reset
+    , start
+    , state
+    , stop
+    , total
+    )
+
+
+type State
+    = Ready
+    | Running
+    | Stopped
+
+
+type Stopwatch
+    = Stopwatch
+        { state_ : State
+        , currentLap_ : Int
+        , previousLaps_ : List Int
+        }
+
+
+new : Stopwatch
+new =
+    Stopwatch { state_ = Ready, currentLap_ = 0, previousLaps_ = [] }
+
+
+state : Stopwatch -> State
+state (Stopwatch { state_ }) =
+    state_
+
+
+currentLap : Stopwatch -> String
+currentLap (Stopwatch { currentLap_ }) =
+    formatTime currentLap_
+
+
+previousLaps : Stopwatch -> List String
+previousLaps (Stopwatch { previousLaps_ }) =
+    previousLaps_
+        |> List.reverse
+        |> List.map formatTime
+
+
+advanceTime : String -> Stopwatch -> Stopwatch
+advanceTime time (Stopwatch ({ state_, currentLap_ } as stopwatch)) =
+    case state_ of
+        Running ->
+            Stopwatch { stopwatch | currentLap_ = currentLap_ + parseTime time }
+
+        _ ->
+            Stopwatch stopwatch
+
+
+total : Stopwatch -> String
+total (Stopwatch { currentLap_, previousLaps_ }) =
+    (currentLap_ :: previousLaps_)
+        |> List.sum
+        |> formatTime
+
+
+start : Stopwatch -> Result String Stopwatch
+start (Stopwatch ({ state_ } as stopwatch)) =
+    case state_ of
+        Running ->
+            Err "cannot start an already running stopwatch"
+
+        _ ->
+            Ok (Stopwatch { stopwatch | state_ = Running })
+
+
+stop : Stopwatch -> Result String Stopwatch
+stop (Stopwatch ({ state_ } as stopwatch)) =
+    case state_ of
+        Running ->
+            Ok (Stopwatch { stopwatch | state_ = Stopped })
+
+        _ ->
+            Err "cannot stop a stopwatch that is not running"
+
+
+lap : Stopwatch -> Result String Stopwatch
+lap (Stopwatch ({ state_, currentLap_, previousLaps_ } as stopwatch)) =
+    case state_ of
+        Running ->
+            Ok (Stopwatch { stopwatch | currentLap_ = 0, previousLaps_ = currentLap_ :: previousLaps_ })
+
+        _ ->
+            Err "cannot lap a stopwatch that is not running"
+
+
+reset : Stopwatch -> Result String Stopwatch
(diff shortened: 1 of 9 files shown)


exercises/practice/split-second-stopwatch/tests/Tests.elm
```diff
@@ -0,0 +1,373 @@
...
+                    ]
+                        |> Expect.equalLists
+                            [ Ok (CurrentLap "01:23:45")
+                            , Ok (PreviousLaps [ "01:23:45" ])
+                            , Ok (CurrentLap "04:01:40")
+                            , Ok (Total "05:25:25")
+                            , Ok (PreviousLaps [ "01:23:45", "04:01:40" ])
+                            , Ok (CurrentLap "08:43:05")
+                            , Ok (Total "14:08:30")
+                            , Ok (PreviousLaps [ "01:23:45", "04:01:40", "08:43:05" ])
```
jiegillet, line 371: The `Output` type introduced on top is only used for this exercise, it's the only one where we need to assert the values of different types. It's a little verbose, but it's the best I could come up with.

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
```diff
@@ -0,0 +1,141 @@
...
+
+type State
+    = Ready
+    | Running
+    | Stopped
+
+
+type Stopwatch
+    = Stopwatch
+        { state_ : State
```
ceddlyburge, line 25: what are the underscores for?

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
```diff
@@ -0,0 +1,141 @@
...
+
+total : Stopwatch -> String
+total (Stopwatch { currentLap_, previousLaps_ }) =
+    (currentLap_ :: previousLaps_)
+        |> List.sum
+        |> formatTime
+
+
+start : Stopwatch -> Result String Stopwatch
+start (Stopwatch ({ state_ } as stopwatch)) =
```
ceddlyburge, line 71: Maybe we could use a finite state machine here, instead of using Result (and in the other places)?

exercises/practice/split-second-stopwatch/tests/Tests.elm
ceddlyburge: This is going to make any test failures hard to read. Not sure what the other solutions are, but maybe asserting on all the fields all the time, instead of picking out one for each assertion?

ceddlyburge (approved):

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
jiegillet: Because `state` is an exposed function, so we are not allowed to shadow it. I was a bit lazy with the `_` notation, that's the kind of thing you see in Haskell, but I didn't want to be too verbose.
Would you have a better suggestion?

exercises/practice/split-second-stopwatch/tests/Tests.elm
jiegillet: Here is an example of a test failure (I changed `total` to add 10 seconds)
```
↓ SplitSecondStopwatch
✗ supports very long laps

    Ok (CurrentLap "01:23:45"), Ok (PreviousLaps ["01:23:45"]), Ok (CurrentLap "04:01:40"), Ok (Total "05:25:35"), Ok (PreviousLaps ["01:23:45","04:01:40"]), Ok (CurrentLap "08:43:05"), Ok (Total "14:08:40"), Ok (PreviousLaps ["01:23:45","04:01:40","08:43:05"])
    ╷
    │ Expect.equalLists
    ╵
    Ok (CurrentLap "01:23:45"), Ok (PreviousLaps ["01:23:45"]), Ok (CurrentLap "04:01:40"), Ok (Total "05:25:25"), Ok (PreviousLaps ["01:23:45","04:01:40"]), Ok (CurrentLap "08:43:05"), Ok (Total "14:08:30"), Ok (PreviousLaps ["01:23:45","04:01:40","08:43:05"])
    
    The first diff is at index 3: it was `Ok (Total "05:25:35")`, but `Ok (Total "05:25:25")` was expected.
```

I think it doesn't read to bad, it's clear where the first problem is, and it's very clear that it's about `total`. What do you think?

> maybe asserting on all the fields all the time, instead of picking out one for each assertion?

I was following the problem specs as close as I could, they are only asserting those specific values, it's easier to maintain in the long run.
Also, if I wanted to assert on all the fields, it would not be possible to treat this as an opaque type problem anymore, we would have to expose the inner structure.

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
jiegillet: I was trying to stick to the specs.
How would you write a finite state machine though? With phantom types? Or just by ignoring error instructions?

ceddlyburge: I've seen `aState` used before, which fits with it being an instance of the state type

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
ceddlyburge: I've seen aState used before, which fits with it being an instance of the state type

exercises/practice/split-second-stopwatch/tests/Tests.elm
ceddlyburge: It doesn't read too bad, but the types don't relate to anything in the code, which is confusing. Maybe using strings would fix that?
Following the problem specs makes sense.

exercises/practice/split-second-stopwatch/.meta/src/SplitSecondStopwatch.example.elm
jiegillet: I don't feel that strongly about it, but you could check what other people think. I suspect it won't be that easy to change it, and they might suggest to create a new exercise based on finite state machines (which would be cool).
In the mean time, I'll keep it this way.

jiegillet:
Thank you for the review!

PR #750, opened 2025-05-16
Add forked concept `recursion` and concept exercise `bird-count`

jiegillet:
Part of #692

I took this concept and exercise combo pretty much straight out of the Elixir track, which itself forked it from the C# track.

Some notes:
- The original issue include teaching about tail call recursion, but since Elixir was separating the two, I thought it would be easier to do the same, so this PR does not talk about tail call recursion.
- I usually leave the `introduction.md` as TODO, but this time, I followed the lead of Elixir that includes more details in the `about.md` and less in the intro. (it's still missing in the exercise though, so the CI will fail)
- I again used AI to help me translate from Elixir to Elm, very useful
- I added a bunch of exercises to practice the concept. There are definitely more that do, but I didn't go through all of them
- I removed the fields `strings` and `lists` from some exercises that practiced those, because we had to many, (`bin/configlet lint` warns that we should have at most 10, now we are OK)

### DIFF
--- a/concepts/recursion/about.md
+++ b/concepts/recursion/about.md
@@ -37,6 +37,8 @@ fibonacci n =
             fibonacci (n - 1) + fibonacci (n - 2)
 ```
 
+Note that this particular implementation is exponentially slow, since related Fibonacci numbers are being calculated from scratch in two independent recursive calls.
+
 Counting the number of occurrences of some given value `x` in a list has two recursive cases:
 
 ```elm
--- a/concepts/recursion/about.md
+++ b/concepts/recursion/about.md
@@ -85,9 +85,9 @@ Under the hood, these functions are implemented using recursion.
 
 ## Infinite execution
 
-Recursive functions, if implemented incorrectly, might never return their result.
+Recursive functions, if implemented incorrectly or for a number of recursion calls too large, might never return their result.
 This can be problematic because each time a function is called, a reference is stored in memory where the runtime should return the result (on the [call stack][wiki-call-stack]).
-If a recursive function calls itself infinitely, it is possible to run out of memory causing the runtime to crash (a [stack overflow error][wiki-stack-overflow]).
+If a recursive function calls itself too much or even infinitely, it is possible to run out of memory causing the runtime to crash (a [stack overflow error][wiki-stack-overflow]).
 Elm's runtime is optimized for recursion and reliability, but infinite recursion can still cause issues.
 
 This problem of infinite execution can be caused by:
--- a/concepts/recursion/about.md
+++ b/concepts/recursion/about.md
@@ -58,7 +58,7 @@ countOccurrences list x =
 Due to immutability, loops in Elm are written differently from imperative languages.
 For example, loops commonly look like:
 
-```
+```c
 for(i = 0; i < array.length; i++) {
   // do something with array[i]
 }
--- a/concepts/recursion/about.md
+++ b/concepts/recursion/about.md
@@ -1,0 +1,102 @@
+# About
+
+Recursive functions are functions that call themselves.
+
+A recursive function needs to have at least one _base case_ and at least one _recursive case_.
+
+A _base case_ returns a value without calling the function again.
+A _recursive case_ calls the function again, modifying the input so that it will at some point match the base case.
+
+In Elm, recursive functions are typically defined using pattern matching with `case` expressions.
+
+```elm
+count : List a -> Int
+count list =
+    case list of
+        [] ->
+            0  -- base case
+
+        _ :: tail ->
+            1 + count tail  -- recursive case
+```
+
+A recursive function can have many base cases and/or many recursive cases.
+For example [the Fibonacci sequence][fibonacci] is a recursive sequence with two base cases:
+
+```elm
+fibonacci : Int -> Int
+fibonacci n =
+    case n of
+        0 ->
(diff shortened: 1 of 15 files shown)


concepts/recursion/about.md
```diff
@@ -0,0 +1,102 @@
...
+            else
+                countOccurrences tail x
+```
+
+## Loops through recursion
+
+Due to immutability, loops in Elm are written differently from imperative languages.
+For example, loops commonly look like:
+
+```
```
ceddlyburge, line 61: Maybe mark this as c or something?

concepts/recursion/about.md
```diff
@@ -0,0 +1,102 @@
...
+    case list of
+        [] ->
+            0  -- base case
+
+        _ :: tail ->
+            1 + count tail  -- recursive case
+```
+
+A recursive function can have many base cases and/or many recursive cases.
+For example [the Fibonacci sequence][fibonacci] is a recursive sequence with two base cases:
```
ceddlyburge, line 24: I think this is a famously slow way to solve the fibonacci sequence. Probably worth mentioning that or choosing a different example?

concepts/recursion/about.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+In practice, iterating over lists and other data structures in Elm is most often done using functions from the [`List`][module-list] module, such as `List.map`, `List.filter`, and `List.foldl`.
+Under the hood, these functions are implemented using recursion.
+
+## Infinite execution
+
+Recursive functions, if implemented incorrectly, might never return their result.
+This can be problematic because each time a function is called, a reference is stored in memory where the runtime should return the result (on the [call stack][wiki-call-stack]).
+If a recursive function calls itself infinitely, it is possible to run out of memory causing the runtime to crash (a [stack overflow error][wiki-stack-overflow]).
```
ceddlyburge, line 90: This can also happen for large / inefficient cases, but can't happen if tail call elimination happens.

exercises/concept/bird-count/.docs/hints.md
```diff
@@ -0,0 +1,41 @@
+# Hints
+
+## General
+
+- Read about recursion [here][recursion].
```
ceddlyburge, line 5: Using 'here' for links is an antipattern.
```suggestion
- Read about [recursion from the creator of Elm][recursion].
```

exercises/concept/bird-count/.docs/instructions.md
```diff
@@ -0,0 +1,71 @@
+# Instructions
+
+You're an avid bird watcher that keeps track of how many birds have visited your garden on any given day.
+
+You decided to bring your bird watching to a new level and implement a few tools that will help you track and process the data.
+
+You have chosen to store the data as a list of integers.
+The first number in the list is the number of birds that visited your garden today, the second yesterday, and so on.
+
+```exercism/note
```
ceddlyburge, line 10: This is a shame, is there not an example of using recursion where it would be the idiomatic thing to do?

exercises/concept/bird-count/.meta/Exemplar.elm
```diff
@@ -0,0 +1,58 @@
+module BirdCount exposing (busyDays, hasDayWithoutBirds, incrementDayCount, today, total)
+
+
+today : List Int -> Maybe Int
+today counts =
+    case counts of
```
ceddlyburge, line 6: I would use List.head here

ceddlyburge (approved):

concepts/recursion/about.md
jiegillet: It's not C, because of `array.length`. I'll mention it's pseudocode

concepts/recursion/about.md
jiegillet: I reworked the text to mention large cases, but I want to keep tail call elimination for the next concept.
That being said, for infinite recursion, tail call optimization can be a negative thing, because it won't overflow, it'll keep heating up your CPU for no benefit :)

concepts/recursion/about.md
jiegillet: I couldn't find a different example that was as simple as this one and not terribly slow, so I added a note.

exercises/concept/bird-count/.docs/instructions.md
jiegillet: Yes, I agree that it's a shame, however I generally think that it's a good exercise to re-implement core functions to deepen your understanding of them. With the understanding that it's not idiomatic of course, and in that sense, a concept exercise is the right place to do it.

Pretty much all practice exercise that are marked with `recursion` are examples where recursion is the idiomatic thing to do.

I do realize that I'm biased because I don't really want to come up with a completely new exercise, but I will:
- tweak the note to emphasize that this is not idiomatic, but for the sake of learning
- come up with an exercise for `tail-call-recursion` that uses recursion idiomatically

exercises/concept/bird-count/.meta/Exemplar.elm
jiegillet: Right, but all these are pretty much one liners if you use `List`, except for `incrementDayCount`.

```elm
today = List.head
hasDayWithoutBirds = List.any ((==) 0)
total = List.sum
busyDays = List.filter ((>=) 5) >> List.length
```

PR #751, opened 2025-05-16
Add concept `tail-call-recursion`

### LINKED ISSUE #692: Concept: recursion

jiegillet:
Closes #692 

~~I only added the concept so far, and now I'm questioning whether this should be part of #750, because any exercise requiring recursion would work out fine, maybe with an extra task "make it tail recursive".
Or maybe I'm just being lazy and I'll come up with an exercise idea later, let's see.~~

The beginning of the introduction was copied from the Elixir track.

Note: [this article](https://jfmengels.net/tail-call-optimization/) was very useful in creating this concept, and [the elm-review rule](https://package.elm-lang.org/packages/jfmengels/elm-review-performance/latest/) it introduces will most likely be used in the analyzer.

### DIFF
--- a/concepts/tail-call-recursion/about.md
+++ b/concepts/tail-call-recursion/about.md
@@ -1,0 +1,64 @@
+# About
+
+A function is [tail-recursive][recursion-tc] if the _last_ thing executed by the function is a call to itself.
+
+Each time any function is called, a _stack frame_ with its local variables and arguments is put on top of [the function call stack][call-stack].
+When a function returns, the stack frame is removed from the stack.
+
+Tail-recursive functions allow for [tail call optimization][tail-call-optimization] (or tail call elimination).
+It's an optimization that allows reusing the last stack frame by the next function call when the previous function is guaranteed not to need it anymore.
+This mitigates concerns of overflowing the function call stack, which is a situation when there are so many frames on the function call stack that there isn't any memory left to create another one.
+
+## Tail Call Optimization in Elm
+
+Under some condition, the Elm compiler is able to automatically performs an a tail call optimization when compiling to JavaScript.
+
+The optimization can happen for a recursive function when the _last_ operation in a branch is done by calling the function itself in a simple function application.
+Let's look at some examples:
+
+```elm
+factorial : Int -> Int
+factorial n =
+  if n <= 1 then
+    n
+  else
+    n * factorial (n-1)
+```
+
+The implementation above is not tail recursive, because the last operation in the `else` branch is a multiplication `n *`.
+
+```elm
+factorial : Int -> Int
+factorial n =
+  factorialHelper n n
+
+factorialHelper : Int -> Int -> Int
+factorialHelper n resultSoFar =
+  if n <= 1 then
+    resultSoFar
+  else
+    factorialHelper (n-1) (n * resultSoFar)
+```
+
+The implementation above is tail recursive and will be optimized, because the last operation in the `else` branch is `factorialHelper` calling itself.
+This would not be possible for a function with the type signature `Int -> Int`, and in practice tail call optimization is often achieved by defining helper functions.
+
+## Edge cases
+
+In some cases, the compiler is not able to optimize recursive functions, let's look at some examples.
+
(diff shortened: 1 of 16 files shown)


concepts/tail-call-recursion/about.md
```diff
@@ -0,0 +1,64 @@
+# About
+
+A function is [tail-recursive][recursion-tc] if the _last_ thing executed by the function is a call to itself.
+
+Each time any function is called, a _stack frame_ with its local variables and arguments is put on top of [the function call stack][call-stack].
+When a function returns, the stack frame is removed from the stack.
+
+Tail-recursive functions allow for [tail call optimization][tail-call-optimization] (or tail call elimination).
+It's an optimization that allows reusing the last stack frame by the next function call when the previous function is guaranteed not to need it anymore.
+This mitigates concerns of overflowing the function call stack, which is a situation when there are so many frames on the function call stack that there isn't any memory left to create another one.
```
ceddlyburge, line 10:
```suggestion
This mitigates concerns of overflowing the function call stack (stack overflow), which is a situation when there are so many frames on the function call stack that there isn't any memory left to create another one.
```

concepts/tail-call-recursion/about.md
```diff
@@ -0,0 +1,64 @@
+# About
```
ceddlyburge, line 1: This is a great explanation

ceddlyburge:
There is quite a lot of nuance to this, so maybe it is worth its own concept?

concepts/tail-call-recursion/about.md
jiegillet: I'm sitting on the shoulders of giants (@neenjaw)

exercises/practice/pascals-triangle/.meta/src/PascalsTriangle.example.elm
```diff
@@ -11,15 +11,11 @@ nextRow row =
...
             else
-                row :: loop (i + 1) (nextRow row)
+                loop (i + 1) (nextRow row) (row :: rowsSoFar)
     in
-    if n < 0 then
-        []
-
-    else
-        loop 0 [ 1 ]
+    loop 0 [ 1 ] []
```
jiegillet, line 21: I tweaked this to make it tail recursive and add it to the list

jiegillet:
@ceddlyburge I found an exercise where recursion is needed, but it's very mathy. What do you think about it?

ceddlyburge:
Other thoughts from a quick google are binary search, towers of hanoi, search for a file in a folder structure

ceddlyburge:
The bfs function is in the relative-distance exercise that andras recently added.

I think the recursion part of the relative-distance exercise is fairly minor, so I don't think it makes sense as a concept exercise (most of the problem is not about the concept). It would work as a practice exercise I think though.

Which exercise are you talking about?

jiegillet:
> Which exercise are you talking about?

`pipers-pie`, the one I added to this PR :)

jiegillet:
Good! Then if you think the exercise is suitable, can I ask you to review the PR? :)

PR #752, opened 2025-05-18
Add `relative-distance`

### LINKED ISSUE #739: Add additional practice exercises?

BNAndras:
Closes #739. Building up the graph of connections took a while :)

### DIFF
--- a/config.json
+++ b/config.json
@@ -1398,7 +1398,8 @@
           "maybe",
           "lists",
           "set",
-          "pattern-matching"
+          "pattern-matching",
+          "recursion"
         ],
         "difficulty": 5
       },
--- a/config.json
+++ b/config.json
@@ -1391,11 +1391,14 @@
         "name": "Relative Distance",
         "uuid": "9e401593-66c2-45ae-8e8b-c8416711d40f",
         "practices": [
-          "dict",
-          "maybe"
+          "dict"
         ],
         "prerequisites": [
-          "dict"
+          "dict",
+          "maybe",
+          "lists",
+          "set",
+          "pattern-matching"
         ],
         "difficulty": 5
       },
--- a/config.json
+++ b/config.json
@@ -1387,6 +1387,19 @@
         "difficulty": 5
       },
       {
+        "slug": "relative-distance",
+        "name": "Relative Distance",
+        "uuid": "9e401593-66c2-45ae-8e8b-c8416711d40f",
+        "practices": [
+          "dict",
+          "maybe"
+        ],
+        "prerequisites": [
+          "dict"
+        ],
+        "difficulty": 5
+      },
+      {
         "slug": "yacht",
         "name": "Yacht",
         "uuid": "394ef7c1-e4b3-484f-a1a3-754870335423",
--- a/exercises/practice/relative-distance/.docs/instructions.md
+++ b/exercises/practice/relative-distance/.docs/instructions.md
@@ -1,0 +1,39 @@
+# Instructions
+
+Your task is to determine the degree of separation between two individuals in a family tree.
+
+- You will be given an input, with all parent names and their children.
+- Each name is unique, a child _can_ have one or two parents.
+- The degree of separation is defined as the shortest number of connections from one person to another.
+- If two individuals are not connected, return a value that represents "no known relationship."
+  Please see the test cases for the actual implementation.
+
+## Example
+
+Given the following family tree:
+
+```text
+      ┌──────────┐            ┌──────────┐ ┌───────────┐
+      │  Helena  │            │  Erdős   │ │  Shusaku  │
+      └───┬───┬──┘            └─────┬────┘ └──────┬────┘
+      ┌───┘   └───────┐             └──────┬──────┘
+      ▼               ▼                    ▼
+┌──────────┐     ┌────────┐          ┌──────────┐
+│   Isla   │     │  Tariq │          │   Kevin  │
+└────┬─────┘     └────┬───┘          └──────────┘
+     ▼                ▼
+┌─────────┐      ┌────────┐
+│   Uma   │      │ Morphy │
+└─────────┘      └────────┘
+```
+
+The degree of separation between Tariq and Uma is 3 (Tariq → Helena → Isla → Uma).
+There's no known relationship between Isla and [Kevin][six-bacons], as there is no connection in the given data.
+The degree of separation between Uma and Isla is 1.
+
+~~~~exercism/note
+Isla and Tariq are siblings and have a separation of 1.
+Similarly, this implementation would report a separation of 2 from you to your father's brother.
+~~~~
+
+[six-bacons]: https://en.m.wikipedia.org/wiki/Six_Degrees_of_Kevin_Bacon
--- a/exercises/practice/relative-distance/.docs/instructions.md
+++ b/exercises/practice/relative-distance/.docs/instructions.md
@@ -1,6 +1,7 @@
 # Instructions
 
 Your task is to determine the degree of separation between two individuals in a family tree.
+This is similar to the pop culture idea that every Hollywood actor is [within six degrees of Kevin Bacon][six-bacons].
 
 - You will be given an input, with all parent names and their children.
 - Each name is unique, a child _can_ have one or two parents.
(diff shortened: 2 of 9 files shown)


exercises/practice/relative-distance/src/RelativeDistance.elm
```diff
@@ -0,0 +1,8 @@
+module RelativeDistance exposing (degreeOfSeparation)
+
+import Dict exposing (Dict)
+
+
+degreeOfSeparation : Dict String (List String) -> String -> String -> Int
+degreeOfSeparation familyTree personA personB =
+    Debug.todo "implement this function"
```
jiegillet, line 8:
```suggestion
    Debug.todo "implement degreeOfSeparation"
```

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
```diff
@@ -0,0 +1,102 @@
...
+    let
+        addRelationship from to tree =
+            Dict.update from
+                (\maybeLinks ->
+                    case maybeLinks of
+                        Nothing ->
+                            Just [ to ]
+
+                        Just links ->
+                            if List.member to links then
```
jiegillet, line 31: [ignore if lazy] you could use a `Set` for this? Here you only need to use `member` and later you use `foldl`.

You wouldn't even need the `member` here, and I'm also thinking that you could replace the `(List.filter (\c -> c /= child) children)` by a `Set.remove child` at the end.

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
```diff
@@ -0,0 +1,102 @@
...
+        ( current, distance ) :: rest ->
+            if current == target then
+                distance
+
+            else
+                case Dict.get current graph of
+                    Nothing ->
+                        bfs graph rest visited target
+
+                    -- dead end --
```
jiegillet, line 87: What is a dead end? `Just connections`? or `Nothing?`? It's a little confusing.

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
```diff
@@ -0,0 +1,102 @@
...
+            -1
+
+        ( current, distance ) :: rest ->
+            if current == target then
+                distance
+
+            else
+                case Dict.get current graph of
+                    Nothing ->
+                        bfs graph rest visited target
```
jiegillet, line 85: Is this possible for a name not to be in a graph? If not, you might as well return `-1` here, cut the computation short.

config.json
```diff
@@ -1367,6 +1367,19 @@
...
+      {
+        "slug": "relative-distance",
+        "name": "Relative Distance",
+        "uuid": "9e401593-66c2-45ae-8e8b-c8416711d40f",
+        "practices": [
+          "dict",
+          "maybe"
+        ],
+        "prerequisites": [
+          "dict"
```
jiegillet, line 1379: Let's just practice `dict` on this one.
For prerequisites, let's add:
- maybe
- lists
- set
- recursion
- pattern matching

maybe tail-call-recursion once #751 is sorted out?

exercises/practice/relative-distance/tests/Tests.elm
```diff
@@ -0,0 +1,242 @@
...
+                \() ->
+                    RelativeDistance.degreeOfSeparation
+                        (Dict.fromList
+                            [ ( "Priya", [ "Rami" ] )
+                            , ( "Kaito", [ "Elif" ] )
+                            ]
+                        )
+                        "Priya"
+                        "Kaito"
+                        |> Expect.equal -1
```
jiegillet, line 56: Should we return `Nothing` here? It would be a little more idiomatic.

jiegillet (commented):
Oh yeah, I remember writing down that list was a pain :D
Well done, nice work!

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
BNAndras: I did use Sets for a while before I simplified my approach. I can put it back.

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
BNAndras: `Nothing`. `elm-format` moved that comment away from the previous line. It did that to the other comment I already removed.

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
BNAndras: I’m not sure I understand. We already checked on line 13 if persons A & B are in the graph.

config.json
BNAndras: The prerequisites seem fine. However, this seems solvable without knowing sets.

exercises/practice/relative-distance/tests/Tests.elm
BNAndras: Makes perfect sense to me.

exercises/practice/relative-distance/src/RelativeDistance.elm
BNAndras: Other exercises just say implement this function. I prefer the clarity of your recommendation though. Should that be done track-wide?

config.json
jiegillet: Your solution uses sets. And in general, I would suggest for students to know about sets before they try this one, because it can really help.

exercises/practice/relative-distance/.meta/src/RelativeDistance.example.elm
jiegillet: Technically, you check if the persons are in the family tree, not the family graph.
Regardless, what I meant is that I do believe that `Dict.get current graph` should never give you `Nothing` (unless there's a bug), so if you do get it, it's a little strange to carry on the computation, it might be harder to find bugs later on.

exercises/practice/relative-distance/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
...
+│   Uma   │      │ Morphy │
+└─────────┘      └────────┘
+```
+
+The degree of separation between Tariq and Uma is 3 (Tariq → Helena → Isla → Uma).
+There's no known relationship between Isla and [Kevin][six-bacons], as there is no connection in the given data.
+The degree of separation between Uma and Isla is 1.
+
+~~~~exercism/note
+Isla and Tariq are siblings and have a separation of 1.
```
ceddlyburge, line 35: From the text graph above it looks like the separation would be 2? Or am I misunderstanding something?

BNAndras: There's an open problem-specs PR to make the sibling link in the graph more clear. You don't need to go to the shared parent first (Helena).  `The degree of separation between Tariq and Uma is 3 (Tariq → Helena → Isla → Uma).` will also be updated since the shortest path is actually Tariq → Isla → Uma for two degrees.

BNAndras: Actually, that PR was merged this morning so I can pull down the updated instructions.

ceddlyburge: Cool, that makes sense. It does seem like the family tree example should accurately reflect the problem.

jiegillet (approved):
Looks good, thank you again!
If you could merge `main` and add the recursion prerequisite, that would be great!

config.json
BNAndras: Done.

PR #755, opened 2025-06-20
Update Tests.elm

pwadsworth:
Last test expects the first two values of previousLaps flipped from the correct order.


ceddlyburge:
Thanks for the thought anyway:)

PR #757, opened 2025-06-26
Add `flower-field` to replace `minesweeper`

BNAndras:
See https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/17967.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1372,9 +1372,9 @@
         "difficulty": 8
       },
       {
-        "slug": "minesweeper",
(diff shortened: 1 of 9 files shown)


jiegillet (approved):
Flower power 🌷 

Thank you for making the change, I double checked that nothing has changed except for names (bomb -> flower, minefield -> garden), you were very thorough. You even updated the comment in tests.toml.

PR #758, opened 2025-06-27
Add new practice exercise `palindrome-products`

jiegillet:
This is me procrastinating on working on https://github.com/exercism/elm-analyzer/issues/96 because finding recursive functions in is hard :)

This one is potentially very hard, and I'm not sure of what the best approach is.
This exercise is about finding the largest/smallest product `i * j` that are also a palindromes for numbers `i` and `j` integer values within a certain range.
If you go at it naively (listing all possible pairs, finding the palindrome products and keeping the largest ones) two tests in particular will never pass because their range is 1000 to 9999, which is more than 80 million pairs (half if you only consider ordered pairs), which is simply too much for Elm to complete within the test runner time limit (of the order of 10 seconds I think).

If you take a smarter approach, such as generating pairs in descending/ascending order and stopping the search at the first palindrome product, then it's not problem, but it's a lot harder to code. For example, I wanted to use an algorithm requiring the use of a priority queue, so I had to build my own, which was difficult (although I love those kind of challenges).

In the Elixir track, we chose to mark those two test as "slow" and exclude them from the test runner, [with this message](https://github.com/exercism/elixir/blob/main/exercises/practice/palindrome-products/.docs/instructions.append.md) to students. We could do something similar, although I don't like the approach so much, because to me, this exercise is about finding an efficient algorithm. I'm sure most students would try a simplest/brute force approach, see it pass, and move on, which to be honest is fine if our goal is to teach the Elm syntax.

I was leaning towards leaving all the tests in and adding a note saying that this exercise is about finding an efficient algorithm.

@ceddlyburge what's your take on this?

### DIFF
--- a/config.json
+++ b/config.json
@@ -1633,6 +1633,29 @@
           "records"
         ],
         "difficulty": 4
+      },
+      {
+        "slug": "palindrome-products",
+        "name": "Palindrome Products",
+        "uuid": "3ef15a49-8f28-4a27-aa83-695db65177f9",
+        "practices": [
+          "tail-call-recursion"
+        ],
+        "prerequisites": [
+          "recursion",
+          "tail-call-recursion",
+          "lists",
+          "records",
+          "result",
+          "pattern-matching",
+          "custom-types",
+          "comparison",
+          "maybe",
+          "set",
+          "array",
+          "strings"
+        ],
+        "difficulty": 8
       }
     ],
     "foregone": [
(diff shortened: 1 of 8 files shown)


ceddlyburge:
That instruction appendix looks good. I think we have two options. We can go ahead and release it, and then if we get any feedback (which is unlikely I think, as we never get any), then we can make changes. Or we can find a better way to test / know whether the solution is naive or good. This will be time consuming, probably result in extra complication and move us further away from the problem_specs. So I think I would vote for the first option.

PR #765, opened 2025-08-15
Fix syntax issues with Tests.elm

mdepolli:


jiegillet:
(sorry for the slow reply)
Could you elaborate further? What are the issues you're trying to fix?

PR #784, opened 2026-05-25
Format all `instructions.append.md` to start with an H1 (required but ignored) and H2

IsaacG:
See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592. Append files should all have an H1. Including an H2 helps break the flow from the non-append file to the append file so it does not read as the same stream of thought.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.append.md
+++ b/exercises/practice/anagram/.docs/instructions.append.md
@@ -1,5 +1,5 @@
 # Instructions Append
 
-## Implementation
+## Track specific instructions
 
 You must return the anagrams in the same order as they are listed in the candidate words.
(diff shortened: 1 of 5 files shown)


ceddlyburge (approved):

jiegillet:
Can we change all the h2 to "Track specific instructions" or something that indicates this might differ on other tracks? See the discussion [on the Elixir track](https://github.com/exercism/elixir/pull/1633).

IsaacG:
Absolutely. Do you want the "Hints" and "Note" to also say that? I just starting working on that change ... though I'll need to run shortly.

jiegillet:
Yes, all the h2, I find it best to be consistent.

IsaacG:
@jiegillet  Done!

jiegillet (approved):
Thank you!

PR #792, opened 2026-09-12
Update introduction.md

codersteinersky:
Still very much a novice but pretty sure the helper should be called with n 1 to set the accumulator properly. As shown the result is 600 instead of 120 for 5!


mpizenberg:
Ah yep good finding! I don’t remember which place is the one we need to change though, because that error appears in a few different places. Will let the other maintainers chip in.

ceddlyburge (commented):
Good spot! Can you update this in the following places, then we can merge it:
- concepts\tail-call-recursion\about.md
- concepts\tail-call-recursion\introduction.md
- exercises\concept\pipers-pie\.docs\introduction.md
Thanks!
Cedd

